Issue #12, opened 2020-06-26
Permissions and Roles


dennis95stumm:
**Describe the solution you'd like**
Roles and permissions are necessary, to allow user, room management and maybe in feature the management of bbb servers only for user with specific roles. Every request to the backend should be checked if it is permitted to the current user or not. Also hiding of specific views and components depending on the user permissions should be considered.

**Additional context**
For the frontend the [casl-vue](https://github.com/stalniy/casl/tree/master/packages/casl-vue) package could be used. See at the [laravel documentations](https://laravel.com/docs/7.x/authorization) for the authorization in the backend. It also makes sense to save the roles and corresponding permissions in the database so a migration will be necessary. Consider creating a management interface for roles and their corresponding permissions for admins of PILOS.

PR #23, opened 2020-07-22
#10 user management admin

rhrt04:
# Part one of \#10

## Type (Highlight the corresponding type)
- Bugfix
- **Feature**
- Documentation
- Refactoring (e.g. Style updates, Test implementation, etc.)
- Other (please describe):

## Checklist
- [x] Code updated to current masters head
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Tests added for the bugfix or newly implemented feature, describe below why if not
- [x] Changelog is updated
- [x] Documentation of code and features exists

## Changes

- Initial admin pages (users, rooms, records, roles, etc. management).
- Added admin page routes in router.js.
- Added admin i18n files for translation relation to admin pages.
- Added admin directory in component directory for admin components.
- Added admin directory in view directory for admin views and subviews. 
- Updated User Resource file to map more informations
- Added User API Controller (CRUD) under /api/v1.
- Added Invitation Controller (C) under api/v1 to check token validity and to create invitation record.
- Added invitationRegister method on RegisterController under api/v1/auth to create user through invitation method.
- Added register method on RegisterController under api/v1/auth to create user.
- Added StoreUser Request for request validation for store and update.
- Added StoreUserInvitation Request for request validation for register user through invitation method.
- Added StoreInvitation Request for request validation for store.
- Added Policies for UserController and InvitationController.
- Added Invitation Model.
- Added invitation-registration.blade.php under views/emails for register invitation link send through email
- Added InvitationRegister under Mail for register invitation link send through email.
- Added Vue CrudLdapModalComponent to create, update, delete user.
- Added Vue InvitationModalComponent to invite user.

PR #45, opened 2020-09-08
12 permission roles 2

dennis95stumm:
# Fixes \#12

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Added management views for roles


dennis95stumm:
@SamuelWei Some frontend tests must be implemented but you can already review the rest of the code.

SamuelWei:
Currently ldap's roles are pulled once when a user logs in for the first time. To keep the roles in sync, the roles should be pulled on every login. Please add an indicator to the user-role association to determine if the role was assigned automatically or manually. If the role was assigned automatically and the user does not have it in LDAP at the next login, it should be removed.

.env.example
```diff
@@ -72,3 +72,4 @@ MIX_ROOM_NAME_LIMIT=50
 
 DEFAULT_LOGO=/images/logo.svg
 DEFAULT_ROOM_LIMIT=-1
+DEFAULT_PAGE_SIZE=15
```
SamuelWei, line 75: @dennis95stumm @rhrt04 We should agree on one setting name, see https://github.com/THM-Health/PILOS/pull/23/files#diff-cbfd64a28982a1818f2b5f16e7f9d634R76

app/Http/Controllers/api/v1/RoleController.php
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Display a listing of the resource.
+     *
+     * @return AnonymousResourceCollection
+     */
+    public function index(Request $request)
+    {
+        $resource = Role::query();
+
+        if ($request->has('sort_by') && $request->has('sort_direction')) {
```
SamuelWei, line 34: @rhrt04 Please use same query params as @dennis95stumm. I prefer sort_by and sort_direction over sort_by and order_by as it is more clear what each parameter defines

app/Http/Controllers/api/v1/RoleController.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        $role             = new Role;
+        $role->name       = $request->name;
+        $role->room_limit = $request->room_limit;
+        $role->default    = false;
+
+        $role->save();
+        $role->permissions()->sync($request->permissions);
+
+        return response()->noContent(200);
```
SamuelWei, line 64: `return new RoleResource($role);`

app/Http/Controllers/api/v1/RoleController.php
```diff
@@ -0,0 +1,137 @@
...
+        $user = Auth::user();
+
+        if (!($user->hasPermission('settings.manage')
+            && $user->hasPermission('roles.update')
+            && $user->hasPermission('roles.viewAny'))) {
+            $role->permissions()->sync($old_role_permissions);
+
+            return response()->json([
+                'error'   => CustomStatusCodes::ROLE_UPDATE_PERMISSION_LOST,
+                'message' => trans('app.errors.role_update_permission_lost')
```
SamuelWei, line 102: `__('app.errors.role_update_permission_lost')`

app/Http/Resources/Role.php
```diff
@@ -15,11 +15,13 @@ class Role extends JsonResource
...
-            'name'        => $this->name,
-            'default'     => $this->default,
-            'updated_at'  => $this->updated_at,
-            'permissions' => Permission::collection($this->whenLoaded('permissions')),
+            'id'           => $this->id,
+            'name'         => $this->name,
+            'default'      => $this->default,
+            'updated_at'   => $this->updated_at,
+            'permissions'  => Permission::collection($this->whenLoaded('permissions')),
+            'modelName'    => $this->modelName,
```
SamuelWei, line 23: Same as https://github.com/THM-Health/PILOS/pull/23#discussion_r493366488

app/Policies/RolePolicy.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Determine whether the user can update the model.
+     *
+     * @param  \App\User $user
+     * @param  \App\Role $role
+     * @return mixed
+     */
+    public function update(User $user, Role $role)
+    {
+        return $user->can('roles.update') && !$role->default;
```
SamuelWei, line 56: why should the default role be non-editable?

SamuelWei (changes requested):

.env.example
dennis95stumm: @rhrt04 I would prefer `DEFAULT_PAGINATION_PAGE_SIZE` but with the value 15, since 10 is too small i think.

app/Policies/RolePolicy.php
dennis95stumm: Why the default role should be editable? I thought that system defaults that get the permissions assigned by the seeder shouldn't be modifiable. If the roles should be different, than new roles can be created and used instead!

dennis95stumm:
> Currently ldap's roles are pulled once when a user logs in for the first time. To keep the roles in sync, the roles should be pulled on every login. Please add an indicator to the user-role association to determine if the role was assigned automatically or manually. If the role was assigned automatically and the user does not have it in LDAP at the next login, it should be removed.

@SamuelWei But only the link not the role?

SamuelWei:
> > Currently ldap's roles are pulled once when a user logs in for the first time. To keep the roles in sync, the roles should be pulled on every login. Please add an indicator to the user-role association to determine if the role was assigned automatically or manually. If the role was assigned automatically and the user does not have it in LDAP at the next login, it should be removed.
> 
> @SamuelWei But only the link not the role?

Yes, only the link

SamuelWei (approved):

app/Policies/RolePolicy.php
SamuelWei: Sounds good, I you didn't knew your thoughts.

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,294 @@
...
+    <hr>
+
+    <b-form @submit='saveRole'>
+      <b-container fluid>
+        <b-row class='my-1'>
+          <b-col sm='3'>
+            <label for='name'>{{ $t('settings.roles.name') }}</label>
+          </b-col>
+          <b-col sm='9'>
+            <b-form-input :state='errors.name ? false : null' id='name' type='text' v-model='model.name' :disabled='isBusy || viewOnly'></b-form-input>
```
SamuelWei, line 18: Please use these functions or implement them globaly, as we need this again and again: https://github.com/THM-Health/PILOS/blob/master/resources/js/components/Room/SettingsComponent.vue#L326

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,294 @@
...
+            <b-form-invalid-feedback>
```
SamuelWei, line 19: Please use these functions: https://github.com/THM-Health/PILOS/blob/master/resources/js/components/Room/SettingsComponent.vue#L326

SamuelWei (changes requested):

resources/js/views/settings/roles/Index.vue
```diff
@@ -0,0 +1,232 @@
...
+  computed: {
+    /**
+     * Returns array with table fields and their labels.
+     */
+    tableFields () {
+      return [
+        { key: 'id', label: this.$t('settings.roles.id'), sortable: true },
+        { key: 'name', label: this.$t('settings.roles.name'), sortable: true },
+        { key: 'default', label: this.$t('settings.roles.default'), sortable: true },
+        { key: 'actions', label: this.$t('settings.roles.actions') }
```
SamuelWei, line 148: Hide column if user has not permission to do any actions: 
image

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,288 @@
...
+          <b-col sm='9'>
+            <b-form-input :state='fieldState("name")' id='name' type='text' v-model='model.name' :disabled='isBusy || viewOnly'></b-form-input>
+            <b-form-invalid-feedback>
+              {{ fieldError('name') }}
+            </b-form-invalid-feedback>
+          </b-col>
+        </b-row>
+        <b-row class='my-1'>
+          <b-col sm='3'>
+            <label for='room-limit'>{{ $t('settings.roles.roomLimit') }}</label>
```
SamuelWei, line 26: Please add checkboxes to set room-limit -1 = unlimited, or room-limit = null (system default)

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,288 @@
...
+              {{ fieldError('room_limit') }}
+            </b-form-invalid-feedback>
+          </b-col>
+        </b-row>
+        <b-row class='my-1'>
+          <b-col sm='3'>
+            <label for='permissions'>{{ $t('settings.roles.permissions') }}</label>
+          </b-col>
+          <b-col sm='9'>
+            <multiselect v-model='model.permissions'
```
SamuelWei, line 40: I think a grouped list would be better to have a structured overview:
image

resources/lang/en/app.php
```diff
@@ -9,8 +10,11 @@
         'file_not_found' => 'The file could not be found.',
         'no_server_available' => 'Currently there are no servers available.',
         'room_limit_exceeded' => 'Creating room failed! You have reached the max. amount of rooms.',
+        'role_delete_linked_users' => 'The role is linked to users and therefore it can\'t be deleted!',
+        'role_update_permission_lost' => 'The changes of the role would lead to a lost of permissions to view and update roles for you!',
     ],
     'model' => [
         'roles' => 'role'
     ],
+    'save_failed' => 'Persistence of the :model entity failed!',
```
SamuelWei, line 19: Saving of the :model data record failed!

SamuelWei (changes requested):

resources/js/views/settings/roles/View.vue
dennis95stumm: Okay, but this makes pagination for the permissions impossible.

SamuelWei: I think we don't need pagination for a permission list; the amount isn't soo big

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,294 @@
...
+          <b-form-input id='name' type='text' v-model='model.name' :state='fieldState("name")' :disabled='isBusy || viewOnly'></b-form-input>
+          <template slot='invalid-feedback'><div v-html="fieldError('name')"></div></template>
+        </b-form-group>
+        <b-form-group
+          label-cols-sm='3'
+          :label="$t('settings.roles.roomLimit.label')"
+          label-for='room-limit'
+          :state='fieldState("room_limit")'
+        >
+          <b-form-input id='room-limit' type='number' :state='fieldState("room_limit")' v-model='model.room_limit' min='-1' :disabled='isBusy || viewOnly || model.room_limit === null || model.room_limit === -1'></b-form-input>
```
SamuelWei, line 28: image

Creating a new role it is possible to select both checkboxes.
I would recommend a radio with 'System default', 'Unlimited', 'Custom amount'
Only if custom amount is selected the input field is shown below the radio.

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,294 @@
...
+                </template>
+
+                <b-form-checkbox-group
+                  v-model='model.permissions'
+                  :options='permissions[key]'
+                  stacked
+                  text-field='translatedName'
+                  value-field='id'
+                  :disabled='isBusy || viewOnly'
+                  :state="Object.keys(errors).some(error => error === 'permissions' || error.startsWith('permissions.')) ? false : null"
```
SamuelWei, line 71: why are you not using the mixins helper?

resources/js/views/settings/roles/View.vue
```diff
@@ -0,0 +1,294 @@
...
+        Base.error(error, this.$root, error.message);
+      }).finally(() => {
+        this.permissionsLoading = false;
+      });
+    },
+
+    /**
+     * Goes back to the role overview page.
+     */
+    back () {
```
SamuelWei, line 227: Whats  the purpose of this method? Why not do it inline?

SamuelWei (changes requested):

resources/js/views/settings/roles/View.vue
dennis95stumm: Because this is a special case that isn't supported by the mixins helper

resources/js/views/settings/roles/View.vue
dennis95stumm: Done

resources/js/views/settings/roles/View.vue
dennis95stumm: Done

resources/js/views/settings/roles/View.vue
SamuelWei: Please change the order of the radio and input field.

dennis95stumm: Done

SamuelWei (approved):

PR #65, opened 2020-10-03
64 Add usage live and statistical data

### LINKED ISSUE #64: Room and server live data and statistics
For rooms and servers data about the usage should be constantly collected and stored for statistical purposes.

A scheduled task should check the status of each server every minute and save the current amount of users, webcams, meetings, etc. as well as the status online/offline. 

The scheduled task should also log the same details for all meetings and create a meeting history for all rooms.

SamuelWei:
# Fixes #64

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Add live and statistical usage data for rooms and servers (amount of meetings, video and audio users)
- Improve error handling on meeting start and join


app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -119,31 +119,74 @@ public function start(Room $room, StartJoinMeeting $request)
...
+            $meeting              = $room->meetings()->create();
             $meeting->start       = date('Y-m-d H:i:s');
             $meeting->attendeePW  = bin2hex(random_bytes(5));
             $meeting->moderatorPW = bin2hex(random_bytes(5));
             $meeting->save();
 
-            if (!$meeting->startMeeting()) {
-                abort(CustomStatusCodes::ROOM_START_FAILED, __('app.errors.room_start'));
+            // Try to find server to host meeting
+            while (true) {
```
dennis95stumm, line 130: The loop is not the best solution! There are some cases where the meeting gets deleted and the loop runs a second time. In this cases an error will be thrown on line 144.

database/factories/Server.php
```diff
@@ -0,0 +1,16 @@
...
+
+use App\Server;
+use Faker\Generator as Faker;
+
+$factory->define(Server::class, function (Faker $faker) {
+    return [
+        'baseUrl' => 'https://test.notld/bigbluebutton/',
+        'salt' => $faker->sha1,
+        'description' => $faker->text,
+        'status' => 1,
```
dennis95stumm, line 13: The status and offline casted as boolean so use here booleans instead

dennis95stumm (changes requested):

database/migrations/2020_10_03_153401_add_current_stat_to_servers_table.php
```diff
@@ -0,0 +1,33 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddCurrentStatToServersTable extends Migration
+{
+    public function up()
+    {
+        Schema::table('servers', function (Blueprint $table) {
+            $table->integer('participant_count')->nullable();
```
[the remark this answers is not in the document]
SamuelWei: For performance reasons in lists and planed load balancers

database/migrations/2020_10_03_153503_add_current_stat_to_rooms_table.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddCurrentStatToRoomsTable extends Migration
+{
+    public function up()
+    {
+        Schema::table('rooms', function (Blueprint $table) {
+            $table->integer('participant_count')->nullable();
```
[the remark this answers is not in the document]
SamuelWei: For performance reasons in lists and planed load balancers

app/Server.php
```diff
@@ -14,6 +14,31 @@ class Server extends Model
...
     ];
 
+    /**
+     * Find server with the lowest usage
+     * @return Server|null
+     */
+    public static function lowestUsage()
+    {
+        $servers = self::where('status', true)->where('offline', false)->get();
+        $sorted  = $servers->sortBy(function ($server, $key) {
```
dennis95stumm, line 24: This could also be done with a mysql query!

dennis95stumm: Maybe the `usage` should be defined as an acessor?

app/Server.php
```diff
@@ -30,7 +55,7 @@ public function bbb()
      */
     public function getMeetings()
     {
-        if (!$this->status or $this->offline) {
+        if (!$this->status) {
```
dennis95stumm, line 58: Why when the server is offline the meetings should be returned?

app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -120,65 +120,59 @@ public function start(Room $room, StartJoinMeeting $request)
         $meeting = $room->runningMeeting();
```
dennis95stumm, line 120: The username in line 118 is used as the id, but in the comment of the user management pull request the username is not required. So it could be that the username is null and then it would lead to an error.

app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -120,65 +120,59 @@ public function start(Room $room, StartJoinMeeting $request)
...
+            $server = Server::lowestUsage();
+
+            // If no server found, throw error
+            if ($server == null) {
+                abort(CustomStatusCodes::NO_SERVER_AVAILABLE, __('app.errors.no_server_available'));
+            }
+
+            $meeting->server()->associate($server);
+            $meeting->room()->associate($room);
             $meeting->save();
```
dennis95stumm, line 138: Why not save the meeting only if everything is successful?

app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -120,65 +120,59 @@ public function start(Room $room, StartJoinMeeting $request)
...
+                $server->offline = true;
+                $server->save();
+                abort(CustomStatusCodes::ROOM_START_FAILED, __('app.errors.room_start'));
+            }
 
-                // Add found server to the meeting
-                $meeting->server()->associate($server);
-                $meeting->save();
+            // Unkown error with the bbb api, that results in meeting not starting
+            if (!isset($result)) {
```
dennis95stumm, line 153: Is there an example where the result doesn't gets set but no exception gets thrown?

app/Server.php
SamuelWei: Status is manually set (enable/disable) server, whereas offline is just an indication if the server is currently offline. To detect that the server is online again, the same function is called so it needs to respond with fresh data.

app/Server.php
SamuelWei: I changed it to an acessor and an sql query

app/Http/Controllers/api/v1/RoomController.php
SamuelWei: In https://github.com/THM-Health/PILOS/blob/64-live-data-statistics/app/Meeting.php#L85 the id of the meeting is needed.

app/Http/Controllers/api/v1/RoomController.php
SamuelWei: I can't find any cases, so I remove it

app/Console/Commands/BuildHistory.php
```diff
@@ -0,0 +1,118 @@
...
+    protected $description = 'Check server status and capture usage data for live data and statistics';
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // TODO End meetings marked in database as still running but not found on server api calls
```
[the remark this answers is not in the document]
SamuelWei: implemented

app/Server.php
```diff
@@ -18,4 +23,44 @@ public function bbb()
...
+        }
+
+        try {
+            $response = $this->bbb()->getMeetings();
+            if ($response->failed()) {
+                return null;
+            }
+
+            return $response->getMeetings();
+        } catch (\Exception $exception) {
```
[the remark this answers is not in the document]
SamuelWei: changed to RuntimeException, as that is thrown by the api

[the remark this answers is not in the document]
SamuelWei: There are RuntimeExceptions from the getMeetings and Exceptions from the bbb constructor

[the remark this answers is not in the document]
SamuelWei: Added TODO, can be improved once the api supports more diverse exceptions

Issue #10, opened 2020-06-12
User management


SamuelWei:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Admin page & api controller for external user CRUD and LDAP RUD. 

**Describe alternatives you've considered**
None

**Additional context**
Compare to greenlight user management

PR #66, opened 2020-10-12
10 user management v2

dennis95stumm:
# Fixes \#10, \#74

- **Bugfix**
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes
- Added administration of users
- Added profile page for management of the current users profile with previous password confirmation
- Fixed not changing headers in roles index table on locale change


resources/js/views/settings/users/Index.vue
```diff
@@ -0,0 +1,267 @@
...
+        config.params.sort_by = ctx.sortBy;
+        config.params.sort_direction = ctx.sortDesc ? 'desc' : 'asc';
+      }
+
+      if (ctx.filter) {
+        config.params.name = ctx.filter;
+      }
+
+      Base.call('users', config).then(response => {
+        this.perPage = response.data.meta.per_page;
```
SamuelWei, line 211: why don't just?
this.meta = response.data.meta

resources/js/views/settings/users/View.vue
```diff
@@ -0,0 +1,510 @@
...
+            :label="$t('settings.users.firstname')"
+            label-for='firstname'
+            :state='fieldState("firstname")'
+          >
+            <b-form-input
+              id='firstname'
+              type='text'
+              v-model='model.firstname'
+              :state='fieldState("firstname")'
+              :disabled="isBusy || config.type === 'view' || model.authenticator !== 'users'"
```
SamuelWei, line 30: I would like to disable the editing of firstname, lastname and email of some users. Could we add this as a group permission e.g.: users.update_own_attributes

resources/js/views/settings/users/View.vue
```diff
@@ -0,0 +1,510 @@
...
+    saveUser (evt) {
+      if (evt) {
+        evt.preventDefault();
+      }
+
+      this.isBusy = true;
+
+      const config = {
+        method: this.config.id === 'new' ? 'post' : 'put',
+        data: {
```
SamuelWei, line 405: why not?
data: this.model

resources/js/views/settings/users/View.vue
```diff
@@ -0,0 +1,510 @@
...
+          this.roles.forEach(role => {
+            role.$isDisabled = !!this.model.roles.find(selectedRole => selectedRole.id === role.id && selectedRole.automatic);
+          });
+          this.model.roles.forEach(role => {
+            role.$isDisabled = role.automatic;
+          });
+        }
+      }).catch(error => {
+        // If the user wasn't found and it is the current user log him out!
+        if (PermissionService.currentUser && this.config.id === PermissionService.currentUser.id && error.response && error.response.status === env.HTTP_NOT_FOUND) {
```
SamuelWei, line 451: what if user isn't found but is not the current user?

resources/js/views/settings/users/View.vue
```diff
@@ -0,0 +1,510 @@
...
+          this.$store.dispatch('session/logout').then(() => {
+            this.$router.push({ name: 'home' });
+          });
+
+        // handle stale errors
```
SamuelWei, line 456: I think this comment is at the wrong place

resources/lang/de/validation.php
```diff
@@ -140,7 +140,8 @@
         ],
         'room' => [
             'already_member' => 'Der Benutzer ist bereits Mitglied des Raums.'
-        ]
+        ],
+        'password' => 'Das Passwort muss ein Zeichen mindestens aus der folgenden vier Kategorien enthalten: Großbuchstabe (A - Z), Kleinbuchstaben (a - z), Zahl (0 - 9), nicht alphanumerisches Zeichen (zum Beispiel: !, $, #, oder %).'
```
SamuelWei, line 144: Das Passwort muss mindestens ein Zeichen aus jeder der folgenden vier Kategorien enthalten:

resources/lang/en/validation.php
```diff
@@ -141,7 +141,8 @@
         ],
         'room' => [
             'already_member' => 'The user is already member of the room.'
-        ]
+        ],
+        'password' => 'The password must contain at least one character from the following four categories: Uppercase letter (A - Z), lowercase letter (a - z), number (0 - 9), non-alphanumeric character (for example: !, $, #, or %).'
```
SamuelWei, line 145: The password must contain at least one character from each of the following four categories:

SamuelWei (changes requested):

resources/js/views/settings/users/View.vue
dennis95stumm: We could do that. Should this be additionally if the user are database users? In other cases this don't make sense if we can't write to the service back, like it is the case for ldap.

resources/js/views/settings/users/View.vue
dennis95stumm: In this case the user will get an 404 error shown by the global error handler. This is not as critical as if the currentUser can't be found.

app/Http/Resources/User.php
```diff
@@ -39,12 +48,19 @@ public function toArray($request)
...
-            'locale'        => $this->locale,
-            'permissions'   => $this->when($this->withPermissions, $this->permissions),
+            'user_locale'   => $this->locale,
+            'permissions'   => $this->when($this->withPermissions, function () {
+                return $this->permissions;
+            }),
             'model_name'    => $this->model_name,
             'room_limit'    => $this->room_limit,
+            'updated_at'    => $this->updated_at,
+            'roles'         => new RoleCollection($this->whenLoaded('roles')),
```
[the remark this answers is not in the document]
dennis95stumm: Good point!

resources/js/views/settings/users/View.vue
SamuelWei: Yes, only for db users. LDAP user attributes should be updated on each login.

dennis95stumm: This should only be used for the profile page correct?

SamuelWei: Yes, the other user editing pages should only be accessible for users with permission to always edit these attributes. The idea behind my comment was, so no student can change their name to something that can't be used to identify them. The information should always come from a trustful source (admin users or ldap). For some users we could grant this permission based on trusting them.

app/Http/Controllers/api/v1/UserController.php
```diff
@@ -88,7 +88,7 @@ public function store(UserRequest $request)
         $user->refresh();
         $user->load('roles');
```
SamuelWei, line 89: do we need this, if the resouce is called with ->withRoles()

resources/js/router.js
```diff
@@ -22,6 +25,20 @@ export const routes = [
...
+    component: UsersView,
+    props: () => {
+      return {
+        config: {
+          id: store.state.session.currentUser ? store.state.session.currentUser.id : 0,
+          type: 'profile'
+        }
+      };
+    },
+    meta: { requiresAuth: true, passwordConfirmation: true }
```
SamuelWei, line 40: If the user wants to edit the attributes, he could also edit the account via the user manager. The password confirmation isn't required on the other route. There could also be the possibility to bypass this check, as it is just client side.

SamuelWei:
It could also happen that the updated_at attribute of a user is null. In that case editing the user fails. We should find a workaround, see if there are any possibilites for creating users with updated_at null or change the db definition.

resources/js/views/settings/users/Index.vue
```diff
@@ -0,0 +1,263 @@
...
+
+      <template v-slot:emptyfiltered>
+        <i>{{ $t('settings.users.nodataFiltered') }}</i>
+      </template>
+
+      <template v-slot:cell(authenticator)="data">
+        {{ $t(`settings.users.authenticator.${data.item.authenticator}`) }}
+      </template>
+
+      <template v-slot:cell(actions)="data">
```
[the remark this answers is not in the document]
dennis95stumm: If think there is the necessary word missing in the sentence? What would you prefer?

resources/js/store/modules/session.js
```diff
@@ -68,7 +68,7 @@ const mutations = {
 
   setCurrentUser (state, currentUser) {
     state.currentUser = currentUser;
-    PermissionService.setCurrentUser(state.currentUser);
+    PermissionService.setCurrentUser(state.currentUser, false);
```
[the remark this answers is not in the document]
dennis95stumm: On logout when the current user is set on the permission service there shouldn't be fired a recompution of the permissions in the can and cannot components, since they will try to remove components due to missing permissions and in the same time they will be removed since the user isn't authenticated any more, what leads to errors.

resources/js/router.js
dennis95stumm: It is not in client side! There is a request send to the backend where the time of last confirmation, which is stored in the session, gets checked. Depending on that the confirmation is rerequested or not.

app/Http/Controllers/api/v1/UserController.php
```diff
@@ -3,19 +3,160 @@
...
+
+    /**
+     * Display a listing of the resource.
+     *
+     * @param  Request                     $request
+     * @return AnonymousResourceCollection
+     */
+    public function index(Request $request)
+    {
+        $resource = User::without('roles');
```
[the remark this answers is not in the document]
dennis95stumm: This may not be required. But will be faster, since the roles wouldn't be loaded unnecessary!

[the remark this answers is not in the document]
SamuelWei: That is the default behavior (https://laravel.com/docs/8.x/eloquent-relationships#eager-loading)

> When accessing Eloquent relationships as properties, the relationship data is "lazy loaded". This means the relationship data is not actually loaded until you first access the property.

resources/js/views/settings/users/Index.vue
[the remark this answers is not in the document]
SamuelWei: fixed comment

resources/js/router.js
dennis95stumm: I removed the password confirmation and will create a ticket to implement this correctly.

app/Http/Controllers/api/v1/UserController.php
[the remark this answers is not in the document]
dennis95stumm: Okay you're right!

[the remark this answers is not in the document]
dennis95stumm: Removed!

resources/js/views/settings/users/View.vue
dennis95stumm: see 3a4780d

dennis95stumm:
> It could also happen that the updated_at attribute of a user is null. In that case editing the user fails. We should find a workaround, see if there are any possibilites for creating users with updated_at null or change the db definition.

see 9fb3832

dennis95stumm:
Instead of a wiki page for the action columns mixin I'll enhance the mixin documentation!

SamuelWei (approved):

SamuelWei (changes requested):
The non-ldap users are logged in via their email, so no username is required for those users. The email adress only needs to be unique for non-ldap users.

dennis95stumm:
> The non-ldap users are logged in via their email, so no username is required for those users. The email adress only needs to be unique for non-ldap users.

Should the non-ldap users be able to set or edit the username?

SamuelWei:
> > The non-ldap users are logged in via their email, so no username is required for those users. The email adress only needs to be unique for non-ldap users.
> 
> Should the non-ldap users be able to set or edit the username?

They should not have a username, so no edit or set.

dennis95stumm:
FYI: [Here](https://github.com/THM-Health/PILOS/blob/16c83fca498adae73b619d73f03d26ce4828326b/app/Http/Controllers/api/v1/RoomController.php#L118) the username is used as an id for all authenticated users, which would lead to errors in the case the user name wouldn't be filled for non-ldap users!

SamuelWei (approved):

Issue #79, opened 2020-12-01
Loading overlay and reload for forms


dennis95stumm:
**Is your feature request related to a problem? Please describe.**
Currently there are forms on the settings pages and maybe on other pages, that just disable the form fields during load and on errors the form fields can be edited, even if the model was not loaded.

**Describe the solution you'd like**
A overlay on all forms should be added, which shows a loading indicator while the form is busy and on errors during the load of the model a reload overlay gets displayed.

**Additional context**
A possible solution was already implemented in the application settings form in 2a7ee10.

PR #85, opened 2020-12-10
79 add loading error overlay

dennis95stumm:
# Fixes \#79

- Bugfix
- Feature
- Documentation
- **Refactoring (e.g. Style updates, Test implementation, etc.)**
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [ ] Changelog is updated

## Changes
- Added loading overlay for roles view page
- Added error handling during load and persist for roles view page
- Added loading overlay for users view page
- Added error handling during load and persist for users view page
- Added check that only the profile page for the own user can be opened

### DIFF
--- a/resources/js/views/settings/users/View.vue
+++ b/resources/js/views/settings/users/View.vue
@@ -361,7 +361,14 @@ export default {
   mounted () {
     EventBus.$on('currentUserChangedEvent', this.togglePermissionFlags);
 
-    if (!['profile', 'view'].includes(this.config.type)) {
+    if (
+      !['profile', 'view'].includes(this.config.type) ||
+      this.config.id === 'new' ||
+      PermissionService.cannot('editUserRole', {
+        id: this.config.id,
+        model_name: 'User'
+      })
+    ) {
       this.loadRoles();
     }
 
--- a/resources/js/views/settings/users/View.vue
+++ b/resources/js/views/settings/users/View.vue
@@ -361,13 +361,11 @@ export default {
   mounted () {
     EventBus.$on('currentUserChangedEvent', this.togglePermissionFlags);
 
-    if (
-      !['profile', 'view'].includes(this.config.type) ||
-      this.config.id === 'new' ||
-      PermissionService.cannot('editUserRole', {
+    if (this.config.id === 'new' || (
+      PermissionService.can('editUserRole', {
         id: this.config.id,
         model_name: 'User'
-      })
+      }) && this.config.type === 'edit')
     ) {
       this.loadRoles();
     }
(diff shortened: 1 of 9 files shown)


resources/js/views/settings/users/View.vue
```diff
@@ -332,12 +361,25 @@ export default {
   mounted () {
     EventBus.$on('currentUserChangedEvent', this.togglePermissionFlags);
 
-    if (this.config.type !== 'profile') {
+    if (!['profile', 'view'].includes(this.config.type)) {
```
SamuelWei, line 364: you also don't need to load the roles if the user is editing himself (not via the profile but via the user page)

SamuelWei (changes requested):

resources/js/views/settings/users/View.vue
```diff
@@ -447,13 +489,18 @@ export default {
           this.$store.dispatch('session/logout').then(() => {
             this.$router.push({ name: 'home' });
           });
+          Base.error(error, this.$root, error.message);
         } else if (error.response && error.response.status === env.HTTP_UNPROCESSABLE_ENTITY) {
           this.errors = error.response.data.errors;
```
SamuelWei, line 494: if there is an error with the roles field, you could trigger a reload of the roles. I did something similar on my open PR

dennis95stumm: So should I change here something?

resources/js/views/settings/users/View.vue
```diff
@@ -361,7 +361,14 @@ export default {
   mounted () {
     EventBus.$on('currentUserChangedEvent', this.togglePermissionFlags);
 
-    if (!['profile', 'view'].includes(this.config.type)) {
+    if (
```
SamuelWei, line 364: the rule is not working correctly. I found many cases where it was loading the data even if not necessary. E.g. why would you want to load the permissions if the user can't edit them?

SamuelWei (changes requested):

resources/js/views/settings/users/View.vue
SamuelWei: I think this should work
```js
 if (this.config.id === 'new' || (
      PermissionService.can('editUserRole', {
        id: this.config.id,
        model_name: 'User'
      }) && this.config.type === "edit")
    ) {
      this.loadRoles();
    }
```

dennis95stumm: Yeah I've used the wrong method. This should be can!

SamuelWei (approved):

Issue #63, opened 2020-10-03
Room list


SamuelWei:
**Is your feature request related to a problem? Please describe.**
The admin should have an overview of all rooms. 

**Describe the solution you'd like**
The list should show the room type, owner, status (running/not running), room name, stats if the room is running (users, video, audio), the server and serverpool

The list should be search and sortable. The component should be generic enough to be used at other places in the system.
The room settings should get a new attribute, like: listed for all users.
Add a room list search and filter for all users to find rooms with these attributes. (Like the course search in moddle)
The list for all users should have roomtype, room name, owner.

Issue #29, opened 2020-08-04
Implement load balancer


SamuelWei:
**Is your feature request related to a problem? Please describe.**
The system should use a load balancer for creating new meetings and display server status.
Similar to https://github.com/blindsidenetworks/scalelite 

**Describe the solution you'd like**
Use data collected by #64 for each new meeting to calculate the server load based on amount of meetings and estimated network traffic, cpu usage, etc. and use a server specific multiplicator to handle different hardware. 

Use the server with the lowest load of the server pool (#30) the room has selected and try to create meeting. If not possible, try next less loaded server. If all servers are not available throw error. 

**Describe alternatives you've considered**
Just use https://github.com/blindsidenetworks/scalelite

dennis95stumm:
@SamuelWei isn't this already implemented in the RoomController?

SamuelWei:
Yes, it was implemented with #88

PR #88, opened 2020-12-14
30 Add basic server management

### LINKED ISSUE #30: Server management
Create views and controller to show, add, update and delete servers.

Servers should be added to the database via the web interface and group to server pools.

A list should give you an overview of all servers in the system.
The list should also show the online status and amount of meetings. If you click on a server you should see all meetings running on the server at the moment.

Each server should be editable and its status, as well as some in #29 required parameters should be configurable.

**Describe alternatives you've considered**
_none_

SamuelWei:
# Fixes #30

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Unify changelog entries
- Unify column naming for server table
- Bugfix for user tests added by #85
- Simplified server status and offline status, replaced with enum (disabled, online, offline)
- Added timestamp to check for stale errors
- Reset server usage on disabled servers
- Added strength for servers, used by load balancer #29
- Added CRUD api for servers
- Added api to check server connection
- Added index and single view for server

## Other information

The in the issue mentioned list of rooms for each server will be implemented with #63


app/Policies/ServerPolicy.php
```diff
@@ -0,0 +1,75 @@
...
+
+use App\Server;
+use App\User;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class ServerPolicy
+{
+    use HandlesAuthorization;
+
+    public function __construct()
```
dennis95stumm, line 13: Remove the empty constructor!

resources/js/views/settings/servers/View.vue
```diff
@@ -0,0 +1,337 @@
...
+            <template slot='invalid-feedback'><div v-html="fieldError('strength')"></div></template>
+          </b-form-group>
+
+          <b-form-group
+            label-cols-sm='4'
+            :label="$t('settings.servers.disabled')"
+            label-for='disabled'
+            :state='fieldState("disabled")'
+          >
+            <b-form-checkbox id="disabled" v-model="model.disabled" name="check-button" switch :disabled='isBusy || modelLoadingError || viewOnly'>
```
dennis95stumm, line 77: The checkbox stays higher than the label, since it gets not centered. A solution can be found here: https://github.com/THM-Health/PILOS/blob/e6bdd60f9c473ecd4e81b5807f568bb2bbf18669/resources/js/views/settings/users/View.vue#L226

resources/js/views/settings/servers/View.vue
```diff
@@ -0,0 +1,337 @@
...
+      this.staleError = {};
+      this.$refs['stale-server-modal'].hide();
+      this.saveServer();
+    },
+
+    /**
+     * Refreshes the current model with the new passed from the stale error response.
+     */
+    refreshServer () {
+      this.model = this.staleError.new_model;
```
dennis95stumm, line 324: During the refresh the api details gets lost. This is caused by the resource `withApi` method in the backend since this not gets called in the `EnsureModelNotStale` middleware.

tests/Frontend/views/settings/servers/Index.spec.js
```diff
@@ -0,0 +1,762 @@
...
+  document.body.appendChild(container);
+  return container;
+};
+
+const store = new Vuex.Store({
+  modules: {
+    session: {
+      namespaced: true,
+      getters: {
+        settings: () => (setting) => setting === 'pagination_page_size' ? 5 : null
```
dennis95stumm, line 79: This does not make sense since in line 62 the per_page is 15.

tests/Frontend/views/settings/servers/Index.spec.js
```diff
@@ -0,0 +1,762 @@
...
+  });
+
+  afterEach(function () {
+    moxios.uninstall();
+  });
+
+  it('list of servers with pagination gets displayed', function (done) {
+    const oldUser = PermissionService.currentUser;
+
+    PermissionService.setCurrentUser({ permissions: ['settings.manage', 'servers.viewAny'] });
```
dennis95stumm, line 97: This could be done in beforeEach!

tests/Frontend/views/settings/servers/Index.spec.js
```diff
@@ -0,0 +1,762 @@
...
+        return view.vm.$nextTick();
+      }).then(() => {
+        const html = view.findComponent(BTbody).findAllComponents(BTr).at(0).html();
+        expect(html).toContain('Server 01');
+        expect(html).toContain(10);
+        expect(html).toContain(5);
+        expect(html).toContain(2);
+
+        view.destroy();
+        PermissionService.setCurrentUser(oldUser);
```
dennis95stumm, line 125: This could be done in afterEach!

dennis95stumm (changes requested):

resources/js/views/settings/servers/View.vue
SamuelWei: Great catch, I didn't see it at first

app/Console/Commands/BuildHistory.php
```diff
@@ -73,7 +85,12 @@ public function handle()
                 continue;
             }
 
-            $serverStat = new ServerStat();
+            $serverStat                          = new ServerStat();
+            $serverStat->participant_count       = 0;
```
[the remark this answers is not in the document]
SamuelWei: No, the default value is still null and if no meeting exisits, the value is still null and that will be saved in the db

database/migrations/2020_12_17_113850_simplify_server_status.php
```diff
@@ -0,0 +1,64 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class SimplifyServerStatus extends Migration
+{
+    public function up()
+    {
+        Schema::table('servers', function (Blueprint $table) {
+            $table->smallInteger('status')->default(ServerStatus::DISABLED)->change();
```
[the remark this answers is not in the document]
SamuelWei: it must be signed, as status can be -1,0 and 1

resources/js/views/settings/servers/View.vue
SamuelWei: Good catch, it also happens on the user view. I'll work on a fix

tests/Frontend/views/settings/servers/Index.spec.js
SamuelWei: True, will change it

database/migrations/2020_12_17_113850_simplify_server_status.php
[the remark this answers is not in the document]
SamuelWei: It can't be changed to tinyInteger, as it is not a supported type for changes, see: https://laravel.com/docs/7.x/migrations

app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -143,7 +144,8 @@ public function start(Room $room, StartJoinMeeting $request)
             catch (\Exception $exception) {
                 // Remove meeting and set server to offline
                 $meeting->forceDelete();
-                $server->offline = true;
+                $server->status     = ServerStatus::OFFLINE;
```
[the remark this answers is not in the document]
SamuelWei: A updating event is now implemented

app/Console/Commands/BuildHistory.php
```diff
@@ -41,7 +42,15 @@ public function handle()
         $servers = Server::all();
 
         foreach ($servers as $server) {
-            if (!$server->status) {
+            if ($server->status == ServerStatus::DISABLED) {
+                $server->participant_count       = null;
```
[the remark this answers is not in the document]
SamuelWei: I removed it, as the event listener now always clears the current usage data if the server is marked as offline or disabled

app/Http/Controllers/api/v1/ServerController.php
```diff
@@ -0,0 +1,151 @@
...
+     * @throws \Exception
+     */
+    public function destroy(Request $request, Server $server)
+    {
+        /**
+         * TODO
+         * Check if meetings are running
+         * Maybe use soft delete or archive?
+         */
+        $server->delete();
```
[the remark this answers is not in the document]
SamuelWei: This is a good question I don't have an answer to, because I'm not sure what the best is.

app/Http/Requests/ServerRequest.php
```diff
@@ -0,0 +1,26 @@
...
+use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Validation\Rule;
+
+class ServerRequest extends FormRequest
+{
+    public function rules()
+    {
+        $rules = [
+            'description'   => ['required', 'string', 'max:255'],
+            'base_url'      => ['required', 'url', 'string', 'max:255', Rule::unique('servers', 'base_url')],
```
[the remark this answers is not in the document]
SamuelWei: while it would be possible to increase it, I don't see why it would make sense. If the base url is very long, it can lead to url to long to be processed by old browsers trying to join a meeting. We write a migration to change the db fields to string, but I would leave the option open if longer urls or descriptions become necessary in the future. Are you fine with that?

database/migrations/2020_12_17_113850_simplify_server_status.php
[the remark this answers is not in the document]
dennis95stumm: Okay, but the status is custom defined and could be always positive.

app/Http/Controllers/api/v1/ServerController.php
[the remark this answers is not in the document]
SamuelWei: The delete was changed to only allow disabled server without any not ended meetings to be deleted.

app/Meeting.php
```diff
@@ -129,18 +129,13 @@ public function isRunning()
...
     {
         $endParams = new EndMeetingParameters($this->id, $this->moderatorPW);
-        $success   = $this->server->bbb()->endMeeting($endParams)->success();
-        if ($success) {
-            $this->end = date('Y-m-d H:i:s');
-            $this->save();
-        }
-
-        return $success;
+        $this->server->bbb()->endMeeting($endParams)->success();
```
dennis95stumm, line 136: The endMeeting method throws a RuntimeException. Shouldn't this error be handled?

app/Http/Kernel.php
```diff
@@ -45,7 +45,7 @@ class Kernel extends HttpKernel
 
         'api' => [
             EnsureFrontendRequestsAreStateful::class,
-            'throttle:60,1',
+            'throttle:200,1',
```
dennis95stumm, line 48: This would mean a user could try 200 times per minute to login? Please add specific throttling for the login routes!

app/Http/Controllers/api/v1/ServerController.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        $server               = new Server();
+        $server->description  = $request->description;
+        $server->base_url     = $request->base_url;
+        $server->salt         = $request->salt;
+        $server->strength     = $request->strength;
+        $server->status       = $request->disabled ? ServerStatus::DISABLED : ServerStatus::ONLINE;
+
+        // Check if server is online/offline and update usage data
+        $server->updateUsage();
```
dennis95stumm, line 111: This cannot work, since the server does not exisits in the database! This would lead to an error!

app/Server.php
```diff
@@ -93,4 +127,181 @@ public function stats()
...
+        $this->save();
+
+        $query   = $this->meetings()->whereNull('end');
+        $total   = $query->count();
+        $success = 0;
+        foreach ($query->get() as $meeting) {
+            try {
+                $meeting->endMeeting();
+                $success++;
+            } catch (\Exception $exception) {
```
dennis95stumm, line 190: But this would mean that if a meeting couldn't be stopped but the others are running, the other meetings wouldn't be finished?

app/Server.php
```diff
@@ -93,4 +127,181 @@ public function stats()
...
+    }
+
+    /**
+     * Delete Server, only possible if no meetings from this system are running and the server is disabled
+     * @return bool|null
+     * @throws \Exception
+     */
+    public function delete()
+    {
+        if ($this->status != ServerStatus::DISABLED || $this->meetings()->whereNull('end')->count() != 0) {
```
dennis95stumm, line 301: Please use a event for this like for the `updating` event!

resources/js/lang/de/settings.js
```diff
@@ -163,5 +163,62 @@ export default {
...
+    nodataFiltered: 'Für die Suchanfrage wurden keine Server gefunden!',
+
+    delete: {
+      item: 'Server {id} löschen',
+      confirm: 'Wollen Sie den Server {id} wirklich entfernen?',
+      title: 'Server löschen?'
+    },
+
+    offlineReason: {
+      connection: 'Connecting to the server failed.',
```
dennis95stumm, line 220: Shouldn't this be in German?

dennis95stumm (changes requested):

app/Http/Kernel.php
SamuelWei: No, I thought so too at first, but laravel has it's own login throttle so nothing to worry about https://laravel.com/docs/7.x/authentication#login-throttling

app/Meeting.php
SamuelWei: I added @throws \Exception to the phpdocblock, so it can be handled my the calling function. The returned boolean returns the api result

app/Http/Controllers/api/v1/ServerController.php
SamuelWei: It does work, as the updateUsage function has some save() calls

app/Server.php
SamuelWei: The boolean of the $meeting->endMeeting(); isn't checked as it doen't make a difference (either the meeting will be ended or it is already ended). If the api call fails, it is handled like all other failed api calls (see check, create, join, etc.) -> the server is marked as offline.

resources/js/lang/de/settings.js
SamuelWei: Fixed

app/Server.php
SamuelWei: Changed

dennis95stumm (approved):

Issue #89, opened 2020-12-16
Maintenance/Info banner


SamuelWei:
**Is your feature request related to a problem? Please describe.**
Like in greenlight it would be great to have a banner for maintenance or other information.

**Describe the solution you'd like**
Adding application settings for info banner: enabled/disabled, color/bootstrap variant, message (html code)

**Describe alternatives you've considered**
Having a html code message can be dangerous. To allow some basic features, adding an icon, heading, message text and link (each optional) could be an alternative. 

**Additional context**
Example screenshot
![image](https://user-images.githubusercontent.com/4281791/102392947-0720ad80-3fd8-11eb-9ae8-9f3c97c7adaa.png)

SamuelWei:
Similar to #78

dennis95stumm:
@SamuelWei Should this be a banner or a flash message?

SamuelWei:
Just a banner that is always present

PR #94, opened 2021-01-05
89 add maintenance info banner

dennis95stumm:
# Fixes \#89

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes
- Added a application banner that can be modified through the application settings
- Fixed error on logout while the user is on the home route of the application


SamuelWei:
Could we hide the banner section in the settings if the banner is disabled?
Could we keep the banner settings saved even if the banner gets disabled, but the content isn't send to all users if disabled? (That would be great for regular banners that just need small adjustments to be reused)

dennis95stumm:
>     * Could we hide the banner section in the settings if the banner is disabled?
> 
>     * Could we keep the banner settings saved even if the banner gets disabled, but the content isn't send to all users if disabled? (That would be great for regular banners that just need small adjustments to be reused)
> 
>     * Could you please add a note to the icon field, should you use e.g. "fas fa-info-circle" or just "fa-info-cirle"?

Done

resources/js/views/settings/Application.vue
```diff
@@ -256,6 +256,187 @@
...
+              :disabled='isBusy || viewOnly || !loaded'
+              :state='fieldState("banner.message")'
+            ></b-form-textarea>
+
+            <template slot='invalid-feedback'>
+              <div v-html="fieldError('banner.message')"></div>
+            </template>
+          </b-form-group>
+
+          <b-form-group
```
SamuelWei, line 338: image

Could be improve this feature to have the link as a button with the bootstrap varients, a text, link, target?

SamuelWei:
image
Could we make the layout a bit more compact to save some space?

SamuelWei:
image
Due to a css class the button text color sometimes is wrong. Also the link doesn't allow mailto: or relative urls

dennis95stumm:
The laravels basic url validation rule is used for the validation of the link. If you look at the [code](https://github.com/laravel/framework/blob/eff2f054606e8251e59023a6f1a4f0fd773634de/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L1607) a maitlo link is possible. But since there always required two slashes the link must be: mailto://foo@bar.de

resources/js/lang/de/settings.js
```diff
@@ -145,6 +145,23 @@ export default {
...
+      title: 'Banner für Mitteilungen',
+      enabled: 'Anzeigen',
+      bannerTitle: 'Überschrift',
+      message: 'Mitteilung',
+      link: 'Anzuzeigender Link nach der Mitteilung',
+      link_text: 'Linktext',
+      link_target: 'Linkziel',
+      link_style: 'Linkart',
+      icon: 'Icon',
+      iconDescription: 'Die CSS-Klasse des Fontawesome-Icons (z. B. `fas fa-door-open`). Der Icon wird nur angezeigt, wenn ein Titel angegeben wurde.',
```
SamuelWei, line 160: https://www.duden.de/rechtschreibung/Icon Das Icon

resources/js/views/settings/Application.vue
```diff
@@ -345,6 +627,21 @@ export default {
       formData.append('room_limit', this.settings.room_limit);
       formData.append('pagination_page_size', this.settings.pagination_page_size);
       formData.append('own_rooms_pagination_page_size', this.settings.own_rooms_pagination_page_size);
+
+      Object.keys(this.settings.banner).forEach(key => {
+        let val = this.settings.banner[key];
+
+        // Since the FormData always strings boolean and empty values must be
```
SamuelWei, line 634: are you sure about that? https://laravel.com/docs/8.x/validation#rule-boolean

SamuelWei (changes requested):

resources/js/views/settings/Application.vue
dennis95stumm: Yes! Due to the FormData, booleans sent as `'true'` and `'false'` and null values as `'null'`. This is not correct since the nullable Rule in the validation wouldn't work and for the boolean validation also will not work since `'true'` != `true`. Only `'0'` and `'1'` supported as strings!

SamuelWei (approved):

PR #105, opened 2021-01-21
Implement included permissions

### LINKED ISSUE #103: Fix inconsistent permissions and policies
**Describe the bug**
Currently there are a fix broken permissions and policies that should be fixed to have a better user experience.

**To Reproduce**
1. Give a role permissions to viewAny user and to edit user, but not to view user
2. Go to user list
3. See missing view button, but existing edit button
4. Click on edit button
5. See error

**Expected behavior**
Either the user should also have the show permission if the user has the edit permission, or the edit should be missing.

Note: This is just an example of the permission bugs. If you try to create a new user it will fail due to missing role permissions, etc.

SamuelWei:
# Fixes #103 

- **Bugfix**
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Add included permissions
- Separate the general settings menu permission from the application settings permissions
- 
## Other information

This new feature fixes some permission inconsistencies where users where for example unable to view a ressource but able to update or delete. Now higher permissions like update include lower permissions like view. It can also be used to address some issues where related ressources could not be viewed due to missing permissions.


app/Http/Controllers/api/v1/RoleController.php
```diff
@@ -90,9 +90,7 @@ public function update(RoleRequest $request, Role $role)
 
         $user = Auth::user();
 
-        if (!($user->hasPermission('settings.manage')
-            && $user->hasPermission('roles.update')
-            && $user->hasPermission('roles.viewAny'))) {
+        if (!$user->hasPermission('roles.update')) {
```
dennis95stumm, line 93: I would prefer to let the check how it is. Since this is more secure if something in the included permissions wouldn't work!

tests/Feature/api/v1/PermissionTest.php
```diff
@@ -21,7 +21,26 @@ public function testIndex()
...
+        $includedPermission1 = factory(Permission::class)->create();
+        $includedPermission2 = factory(Permission::class)->create();
+
+        Permission::SetupIncludedPermissions($permission->name, [$includedPermission1->name,$includedPermission2->name]);
+
+        $this->getJson(route('api.v1.roles.index'))->assertStatus(401);
+        $this->actingAs($user)->getJson(route('api.v1.permissions.index'))
+            ->assertSuccessful()
+            ->assertJson(['data' => [
+                ['id'=>$permission->id,'name'=>$permission->name,'includedPermissions'=>[$includedPermission1->id,$includedPermission2->id]],
```
dennis95stumm, line 41: Shouldn't we add some recursion checks? Or isn't this necessary for the current implementation?

resources/js/views/settings/roles/View.vue
```diff
@@ -329,6 +383,8 @@ export default {
           }
 
           this.permissions[group].push(permission);
+
+          this.includedPermissionMap[permission.id] = permission.includedPermissions;
```
dennis95stumm, line 387: Shouldn't we use here the set function?

dennis95stumm: https://vuejs.org/v2/guide/reactivity.html#For-Objects

dennis95stumm (changes requested):

database/seeds/RolesAndPermissionsSeeder.php
```diff
@@ -22,37 +21,70 @@ public function run()
         $adminPermissions[] = Permission::firstOrCreate([ 'name' => 'rooms.delete' ])->id;
 
         $adminPermissions[] = Permission::firstOrCreate([ 'name' => 'settings.manage' ])->id;
-        $adminPermissions[] = Permission::firstOrCreate([ 'name' => 'settings.viewAny' ])->id;
-        $adminPermissions[] = Permission::firstOrCreate([ 'name' => 'settings.update' ])->id;
+
+        $adminPermissions[] = Permission::firstOrCreate([ 'name' => 'applicationSettings.viewAny' ])->id;
```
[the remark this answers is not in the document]
SamuelWei: > Renaming in seeds is not the best idea since the previously created permissions will stay in the database!

See line 58+59, it addresses this issue

resources/js/views/settings/roles/View.vue
SamuelWei: Good catch! Took me a while to write a test that would fail with the current solution and works with the correct way.

tests/Feature/api/v1/PermissionTest.php
SamuelWei: What do you mean by recursion checks? I added a test to check if it is a problem if permissions include each other (loops)

dennis95stumm:
Please update to masters head!

Issue #106, opened 2021-01-22
User welcome email on user creation and password reset throught an admin


dennis95stumm:
**Is your feature request related to a problem? Please describe.**
Currently it is only possible that the administrator changes the password and gives that password to the user. Also on the user creation the admin must supply the password to the user.

**Describe the solution you'd like**
The administrator should be able to select between a custom password and password reset on user creation. Also a possibility to reset the password for already created users should be available. But in both cases the email dispatch must be configured properly! Maybe a check should be done and in case of wrong configuration an error should be thrown. The user should get an email with a password reset link. Where the user can reset his password. Depending on the case the content of the email must differ since on creation it is a welcome message and on reset it is a reset message. Additionally a self password reset should be configurable.

PR #107, opened 2021-01-22
106 password reset

dennis95stumm:
# Fixes \#106

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Added self password reset for registered users
- Added possibllity to reset registered users password by an admin
- Added possibillity to generate passwords including a password reset mail for new created users


resources/lang/de.json
```diff
@@ -0,0 +1,14 @@
+{
```
SamuelWei, line 1: Why do you want to mix 'short keys' and 'translation strings as keys'? I think staying with short keys is better, as we don't have to change all language files if we only wan't to fix a spelling mistake in the primary language.

app/Console/Commands/DeleteUnverifiedNewUsers.php
```diff
@@ -0,0 +1,47 @@
...
+    public function handle()
+    {
+        $broker           = Password::broker('new_users');
+        $unverified_users = User::join('password_resets', 'password_resets.email', '=', 'users.email')
+            ->where('initial_password_set', '=', true)
+            ->where('password_resets.created_at', '<', Carbon::now()->subMinutes(config('auth.passwords.new_users.expire')))
+            ->get();
+
+        foreach ($unverified_users as $user) {
+            echo 'Deleting user ' . $user->id;
```
SamuelWei, line 40: Maybe use log files?

resources/lang/de.json
```diff
@@ -0,0 +1,14 @@
...
+  "Reset Password Notification": "Passwort-Rücksetz-Benachrichtigung",
+  "You are receiving this email because we received a password reset request for your account.": "Sie erhalten diese E-Mail, weil eine Anfrage zum Zurücksetzen des Passworts für Ihr Benutzerkonto gestellt wurde.",
+  "Reset Password": "Passwort zurücksetzen",
+  "This password reset link will expire in :count minutes.": "Der Rücksetz-Link wird in :count Minuten ablaufen.",
```
SamuelWei, line 5: Maybe it would be better to give the user the absolute time, as the email sending and receiving can be delayed

resources/lang/de.json
```diff
@@ -0,0 +1,14 @@
...
+  "If you did not request a password reset, no further action is required.": "Wenn Sie keine Anfrage zum Zurücksetzen des Passworts gestellt haben, ist keine weitere Aktion nötig.",
+  "Hello!": "Hallo!",
+  "Regards": "Mit freundlichen Grüßen",
+  "If you’re having trouble clicking the \":actionText\" button, copy and paste the URL below\ninto your web browser:": "Wenn Sie Probleme haben den \":actionText\" Button zu drücken, kopieren Sie die nachfolgende URL und fügen in Ihren Browser ein:"
+  "User account created": "Benutzerkonto wurde erstellt",
+  "You are receiving this email because a new user account was created for your email address.": "Sie erhalten diese E-Mail, weil ein neues Nutzerkonto für Ihre E-Mail-Adresse erstellt wurde.",
+  "Change password": "Passwort ändern",
+  "You have :count minutes to change your password.": "Sie haben :count Mintuen, um Ihr Passwort zu ändern."
```
SamuelWei, line 13: Spelling: **Mintuen**

SamuelWei (changes requested):

resources/lang/de.json
dennis95stumm: You're right, but Laravel uses this style for mails and much other content.

dennis95stumm: So I can't change a lot here!

resources/lang/de.json
dennis95stumm: This is done by Laravel. I could only reimplement the whole mail notification to make those notifications.

app/Notifications/UserWelcome.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed       $notifiable
+     * @return MailMessage
+     */
+    public function toMail($notifiable)
+    {
+        $url = url('/reset_password?') . \Arr::query([
```
SamuelWei, line 65: Maybe add a url attribute so the view can show a new user specify text and not the regular password reset 
image

resources/views/vendor/notifications/email.blade.php
```diff
@@ -0,0 +1,44 @@
+@component('mail::message')
```
SamuelWei, line 1: Would be great if you could overwrite the default laravel logo

app/Notifications/UserWelcome.php
SamuelWei, line 65: Maybe add a url attribute so the view can show a new user specify text and not the regular password reset 
image

SamuelWei (changes requested):

resources/views/vendor/notifications/email.blade.php
SamuelWei: I solved this in c03b1c8

app/Notifications/UserWelcome.php
SamuelWei: I solved this in 8432aad

resources/js/views/settings/users/View.vue
```diff
@@ -116,6 +116,25 @@
...
                 <template slot='invalid-feedback'><div v-html="fieldError('user_locale')"></div></template>
               </b-form-group>
+              <b-form-group
+                label-cols-sm='3'
+                :label="$t('settings.users.timezone')"
+                label-for='timezone'
+                :state='fieldState("timezone")'
+              >
+                <b-form-select
+                  :options='model.timezones'
```
SamuelWei, line 126: This doesn't work for new users as the timezone list is never loaded. I'm already working on a global default timezone setting that can pre-select this timezone field.

SamuelWei (changes requested):

app/Http/Resources/ApplicationSettings.php
```diff
@@ -64,6 +65,7 @@ public function toArray($request)
                     'link_target' => setting('banner.link_target'),
                 ])
             ],
+            'timezones' => timezone_identifiers_list(),
```
SamuelWei, line 68: I'm not sure if this is the best idea. This will increase the size of the settings request from 1,8 kB before to 10,2 kB. Do you think a separate route with just the timezones would be better?

resources/js/views/settings/users/View.vue
```diff
@@ -411,6 +416,7 @@ export default {
         password_confirmation: null,
         user_locale: null,
         bbb_skip_check_audio: false,
+        timezone: null,
```
SamuelWei, line 419: Would be great if we could set the timezone to the default, same with the locale.

SamuelWei (changes requested):

app/Notifications/PasswordReset.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    private $expireDate;
+
+    /**
+     * Create a new notification instance.
+     *
+     * @param string $token
+     * @param Carbon $expireDate
+     */
+    public function __construct(string $token, Carbon $expireDate)
```
SamuelWei, line 40: the parameter expireDate is a bit missleading, as it is just the start date to which the expire time is added to. I'm not even sure if we need it at all.

resources/js/views/settings/users/Index.vue
```diff
@@ -94,6 +94,19 @@
               <i class='fas fa-edit'></i>
             </b-button>
           </can>
+          <can method='resetPassword' :policy='data.item'>
+            <b-button
```
SamuelWei, line 98: I would recommend a confirm modal, so no one resets the password by accident

resources/js/views/settings/users/View.vue
```diff
@@ -116,6 +116,35 @@
...
+                  >
+                    <template v-slot:first>
+                      <b-form-select-option :value="null" disabled>{{ $t('settings.users.timezone') }}</b-form-select-option>
+                    </template>
+                  </b-form-select>
+                  <template slot='invalid-feedback'><div v-html="fieldError('timezone')"></div></template>
+
+                  <b-input-group-append>
+                    <b-button
+                      v-if="timezonesLoadingError"
```
SamuelWei, line 141: Should be also disable the saving button like we do it with the roles?

resources/js/views/settings/Application.vue
```diff
@@ -253,6 +253,53 @@
...
+              :disabled='isBusy || timezonesLoading || timezonesLoadingError || viewOnly || !loaded'
+            >
+              <template v-slot:first>
+                <b-form-select-option :value="null" disabled>{{ $t('settings.application.default_timezone') }}</b-form-select-option>
+              </template>
+            </b-form-select>
+            <template slot='invalid-feedback'><div v-html="fieldError('default_timezone')"></div></template>
+            <b-input-group-append>
+              <b-button
+                v-if="timezonesLoadingError"
```
SamuelWei, line 295: Should be also disable the saving button in case loading of the timezones fail?

SamuelWei (changes requested):

app/Notifications/PasswordReset.php
dennis95stumm: Why we shouldn't need this date? What about the expire time in the email?

SamuelWei: As the expireDate in the constructor = now, you can get remove the parameter and set the expire data = now + expire_minutes... in the toMail method. That way you also don't steal time from the user if the email sending process get's delayed.

SamuelWei: I'm sorry! I misunderstood something.

SamuelWei (approved):

PR #108, opened 2021-02-01
Add room list

### LINKED ISSUE #63: Room list

SamuelWei:
# Fixes #63

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Added setting to room to list room in public room search
- Added settings in room types to give rooms the room search setting
- Added room search with filter for all users
- Added list with all rooms for users with new rooms.viewAll permission
- Added list with all currently running meetings and their usage for users with the new meetings.viewAny permission


resources/js/views/App.vue
```diff
@@ -33,7 +33,18 @@
...
+            <b-nav-item :to="{ name: 'rooms.own_index' }" v-if='isAuthenticated'>{{ $t('rooms.myRooms') }}</b-nav-item>
+            <b-nav-item :to="{ name: 'rooms.index' }" v-if='isAuthenticated'>
+              <can method='viewAll' policy='RoomPolicy'>
+                {{ $t('rooms.allRooms') }}
+              </can>
+              <cannot method='viewAll' policy='RoomPolicy'>
+                {{ $t('rooms.findRooms') }}
+              </cannot>
+            </b-nav-item>
+            <can v-if='isAuthenticated' method='viewAny' policy='MeetingPolicy'>
```
dennis95stumm, line 45: This or the settings navigation leads to errors on logout. I think the v-if='isAuthenticated' cannot be used at the can/cannot components

resources/lang/en/validation.php
```diff
@@ -246,6 +246,7 @@
         'server_pool'                       => 'Server pool',
         'icon_text'                         => 'Icon text',
         'color'                             => 'Color',
+        'allow_listing'                     => 'Room search allowed',
```
dennis95stumm, line 249: What about the listed field?

resources/lang/de/validation.php
```diff
@@ -246,5 +246,6 @@
         'server_pool'                       => 'Serverpool',
         'icon_text'                         => 'Icon Text',
         'color'                             => 'Farbe',
+        'allow_listing'                     => 'Raumsuche erlaubt',
```
dennis95stumm, line 249: What about the listed field?

resources/js/views/meetings/Index.vue
```diff
@@ -0,0 +1,224 @@
...
+        </b-input-group>
+      </b-col>
+    </b-row>
+    <hr>
+
+    <b-overlay  :show="loadingError" >
+
+      <template #overlay>
+        <div class="text-center my-2">
+          <b-spinner v-if="isBusy" ></b-spinner>
```
dennis95stumm, line 24: This does not make sense, since the overlay is only shown if there is an loading error!

resources/js/views/meetings/Index.vue
```diff
@@ -0,0 +1,224 @@
...
+      show-empty
+      :busy.sync='isBusy'
+      :fields='tableFields'
+      :items='fetchMeetings'
+      id='meetings-table'
+      ref='meetings'
+      :current-page='currentPage'>
+
+      <template v-slot:empty>
+        <i>{{ filtered ? $t('meetings.nodataFiltered') : $t('meetings.nodata') }}</i>
```
dennis95stumm, line 47: For filtered there is an extra slot, named `emptyfiltered`

resources/js/views/meetings/Index.vue
```diff
@@ -0,0 +1,224 @@
...
+  },
+
+  data () {
+    return {
+      isBusy: false,
+      loadingError: false,
+      currentPage: undefined,
+      total: undefined,
+      perPage: undefined,
+      actionPermissions: [],
```
dennis95stumm, line 174: This does not make sense here!

resources/js/views/meetings/Index.vue
```diff
@@ -0,0 +1,224 @@
...
+          page: ctx.currentPage
+        }
+      };
+
+      if (ctx.sortBy) {
+        config.params.sort_by = ctx.sortBy;
+        config.params.sort_direction = ctx.sortDesc ? 'desc' : 'asc';
+      }
+
+      this.filtered = !!this.filter;
```
dennis95stumm, line 202: This flag is not necessary if you would use the extra slot!

dennis95stumm (changes requested):

app/Policies/RoomPolicy.php
```diff
@@ -28,6 +28,17 @@ public function viewAny(User $user)
...
 
+    /**
+     * Determine whether the user can view all rooms.
+     *
+     * @param  User $user
+     * @return bool
+     */
+    public function viewAll(User $user)
+    {
+        return $user->can('rooms.viewAll');
```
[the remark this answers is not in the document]
SamuelWei: By definition viewAny checks if the user can view any room (not all rooms). How much each user can see if he called the rooms url is definied by the auth. status, role inside the room, room settings, owner, etc. As part of the refactoring of the room permissions there should be an admin permission to view all rooms independent of the room settings. In this PR the permission is used to show normal users a limited list of rooms and the complete list of all rooms just to the admins.

resources/js/views/meetings/Index.vue
SamuelWei: The loadingError is set to false if the requsted was successfull. During the request the loadingError is still true and isBusy true

resources/js/views/meetings/Index.vue
SamuelWei: that is true, but as we don't use a client side filter function we don't know if there are not results or just no results for the search query. So we need this workaround to show the user a different message if a search query is present.

> Content to display when no items are present in the filtered `items` array. Optionally scoped

package.json
```diff
@@ -68,6 +68,7 @@
         "bootstrap-vue": "^2.15.0",
         "es6-promise": "^4.2.8",
         "v-clipboard": "^2.2.3",
+        "vue-dayjs-plugin": "^1.0.0",
```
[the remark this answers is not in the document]
SamuelWei: I replaced it

resources/js/views/meetings/Index.vue
dennis95stumm: But in this case there are two spinner one in the table and one in the overlay!
grafik

resources/js/views/meetings/Index.vue
dennis95stumm: The problem is that you just not set the filter property on the table. In that case the @change-listener in the filter input field wouldn't be necessary, because the function for getting the data would be triggered automatically!

dennis95stumm: This is already used in the users index table for example!

resources/js/views/meetings/Index.vue
dennis95stumm: see comment above!

dennis95stumm (changes requested):
See comments above!

resources/js/views/meetings/Index.vue
SamuelWei: Fixed. I did not use the debounce feature, but lazy to send less requests to the server

dennis95stumm (approved):

Issue #119, opened 2021-03-05
System wide default presentation


dennis95stumm:
**Is your feature request related to a problem? Please describe.**
Currently if no presentation is selected for a room, the default presentation of the installed big blue button is selected and for each server the default presentation must be uploaded.

**Describe the solution you'd like**
Add in the application settings a possibility to upload a system wide default presentation, which should be uploaded to the bbb server if no other presentation is selected for the meeting.

PR #120, opened 2021-03-05
#119 Add default presentation

dennis95stumm:
# Fixes \#119

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Added a configurable system wide default presentation


app/Http/Controllers/api/v1/ApplicationController.php
```diff
@@ -75,4 +87,24 @@ public function currentUser()
...
         return (new UserResource(Auth::user()))->withPermissions();
     }
+
+    /**
+     * Returns the default presentation or 404 if there is no default presentation
+     * or the file is missing.
+     *
+     * @return StreamedResponse
+     */
+    public function defaultPresentation(): StreamedResponse
```
SamuelWei, line 97: Could we somehow use this https://laravel.com/docs/8.x/filesystem#the-public-disk so we don't have to stream the default_presentation via php and can just use the webserver.

Only for the private presentations we have to use the Storage::download to ensure the access is restricted.

SamuelWei (changes requested):

resources/js/views/settings/Application.vue
```diff
@@ -166,139 +166,212 @@
...
+                id='default_presentation'
+                :state="fieldState('default_presentation')"
+                :browse-text="$t('app.browse')"
+                :placeholder="$t('settings.application.defaultPresentation')"
+                v-model="default_presentation"
+              >
+              </b-form-file>
+              <b-form-text v-if="!viewOnly">{{ $t('rooms.files.formats', { formats: settings.bbb.file_mimes }) }}<br>{{ $t('rooms.files.size', { size: settings.bbb.max_filesize }) }}</b-form-text>
+              <template slot='invalid-feedback' v-if="!viewOnly">
+                <div v-html="fieldError('default_presentation')"></div>
```
SamuelWei, line 370: missing translation of the form validator

resources/js/views/settings/Application.vue
```diff
@@ -166,139 +166,212 @@
...
+            </b-form-group>
+          </b-col>
+        </b-row>
+
+        <b-row cols='12'>
+          <b-col md='6'>
+            <b-form-group
+              label-class="font-weight-bold"
+              class="mb-4"
+              :label="$t('settings.application.default_timezone')"
```
SamuelWei, line 289: wrong naming convention. We don't have it equal through the whole app, but let's try to at least use the same standard in a component/view.

resources/js/views/settings/Application.vue
```diff
@@ -166,139 +166,212 @@
...
-          >
-          </b-form-input>
-
-          <template slot='invalid-feedback'>
-            <div v-html="fieldError('pagination_page_size')"></div>
-          </template>
-        </b-form-group>
+        <b-row cols='12'>
+          <b-col md='6'>
+            <b-form-group
```
SamuelWei, line 171: I would prefer a heading like all the other settings 'block's:
image

SamuelWei (changes requested):

app/Meeting.php
```diff
@@ -102,6 +102,10 @@ public function startMeeting()
             $meetingParams->addPresentation($file->getDownloadLink(), null, preg_replace("/[^A-Za-z0-9.-_\(\)]/", '', $file->filename));
         }
 
+        if (empty($meetingParams->getPresentations()) && !empty(setting('default_presentation'))) {
+            $meetingParams->addPresentation(setting('default_presentation'));
```
SamuelWei, line 106: I would prefer a presentation name, so BBB does not use this random name. We could hardcode "default.pdf", use a translated string or a custom variable. @dennis95stumm What do you think?

dennis95stumm: I've hardcoded the name with the extension from the client!

SamuelWei: That's a great solution, I did not even think about the extension

PR #130, opened 2021-04-20
Fix user room count

### LINKED ISSUE #129: Room limit display search
**Describe the bug**
Similar to the bug in #69 the count of the users rooms can somestimes be incorrect.
An incorrect count results in showing the user the modal to create new rooms, even though this is not possible for the user and will result in an error.

**To Reproduce**
1. Go to My Rooms
2. Create rooms until your limit is exceeded
2. See limit is reached like 5/5
3. Enter a random search query to the search field
4. See the displayed count/limit change to e.g. 0/5 and a visible create new room button

**Expected behavior**
The room count should be independent from the query.

SamuelWei:
# Fixes #129 

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [ ] Changelog is updated

## Changes

- Added new meta parameter to display amount of user rooms indipenent from search queries

### DIFF
--- a/tests/Feature/api/v1/Room/RoomTest.php
+++ b/tests/Feature/api/v1/Room/RoomTest.php
@@ -324,12 +324,16 @@ class RoomTest extends TestCase
         $this->actingAs($this->user)->getJson(route('api.v1.rooms.index').'?filter=own')
             ->assertOk()
             ->assertJsonFragment(['id'=>$rooms[2]->id])
-            ->assertJsonFragment(['id'=>$rooms[3]->id]);
+            ->assertJsonFragment(['id'=>$rooms[3]->id])
+            ->assertJsonPath('meta.total', 2)
+            ->assertJsonPath('meta.total_no_filter', 2);
 
         $this->actingAs($this->user)->getJson(route('api.v1.rooms.index').'?filter=shared')
             ->assertOk()
             ->assertJsonFragment(['id'=>$rooms[0]->id])
-            ->assertJsonFragment(['id'=>$rooms[1]->id]);
+            ->assertJsonFragment(['id'=>$rooms[1]->id])
+            ->assertJsonPath('meta.total', 2)
+            ->assertJsonPath('meta.total_no_filter', 2);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


tests/Feature/api/v1/Room/RoomTest.php
```diff
@@ -346,27 +350,37 @@ public function testRoomSearch()
...
+            ->assertJsonPath('meta.total', 1)
+            ->assertJsonPath('meta.total_no_filter', 1);
 
         // Testing with invalid name
         $this->actingAs($this->user)->getJson(route('api.v1.rooms.index').'?filter=own&search=Meeting Two')
             ->assertOk()
-            ->assertJsonMissing(['id'=>$room->id]);
+            ->assertJsonMissing(['id'=>$room->id])
+            ->assertJsonPath('meta.total', 0)
+            ->assertJsonPath('meta.total_no_filter', 1);
```
dennis95stumm, line 383: It would be nice to have a test where no filter is set, to ensure the meta is not set!

tests/Frontend/views/rooms/OwnIndex.spec.js
```diff
@@ -355,14 +364,188 @@ describe('Own Room Index', function () {
...
+        // check empty list message for no user rooms
+        const searchField = view.findComponent({ ref: 'search' });
+        await searchField.setValue('test2');
+        searchField.trigger('change');
+        moxios.requests.reset();
+        moxios.wait(async () => {
+          const ownRequest = moxios.requests.at(0);
+          const sharedRequest = moxios.requests.at(1);
+          expect(ownRequest.url).toBe('/api/v1/rooms?filter=own&page=1&search=test2');
+          expect(sharedRequest.url).toBe('/api/v1/rooms?filter=shared&page=1&search=test2');
```
dennis95stumm, line 421: A test without a filter an the meta would be nice, if it is not already done with the other existing tests!

dennis95stumm (changes requested):

tests/Frontend/views/rooms/OwnIndex.spec.js
SamuelWei: The own index page only uses request with the two filters, the non-filter requests are used for the public room list

tests/Feature/api/v1/Room/RoomTest.php
SamuelWei: As there is no assertJsonPathMissing I added a count test of the meta object attributes to the filter and non-filter request

PR #132, opened 2021-04-22
Error handling room view

### LINKED ISSUE #131: Error handling for click on room
**Describe the bug**
On the user's room overview page, all rooms are displayed as tiles. The user can click on a tile to open the room. If a server error occurs during loading, the error is not reported correctly.

**Expected behavior**
The error should be handled by the global error handler.

SamuelWei:
# Fixes #131 

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Fix broken error handler
- Add 503 error to error handler for laravel maintenance mode


resources/js/api/base.js
```diff
@@ -48,6 +48,8 @@ export default {
       vm.$router.replace({ name: 'home' });
     } else if (responseStatus === env.HTTP_PAYLOAD_TOO_LARGE) { // 413 => payload to large
       vm.flashMessage.error(vm.$t('app.flash.tooLarge'));
+    } else if (responseStatus === env.HTTP_SERVICE_UNAVAILABLE) { // 503 => maintenance mode
+      window.location.reload();
```
dennis95stumm, line 52: Would it be better to delay the reload? Otherwise to many requests would be send.

dennis95stumm (changes requested):

resources/js/api/base.js
dennis95stumm: It would be better to have a maintenance page for this case!

dennis95stumm: There is no message just a page reload, this is not intuitive!

dennis95stumm: On other errors that not handled here, the page will be broken, since the room card will have an overlay with a loading indicator!

dennis95stumm (changes requested):

Issue #143, opened 2021-05-31
Import from newer greenlight versions fails


SamuelWei:
**Describe the bug**
Newer greenlight versions use 4 block  (eg. 123-abc-def-456), instead of 3 block urls (eg. 123-abc-def)
Our database room_id is only 11 chars long.

**To Reproduce**
1. Run import from greenlight cli command with new greenlight versions
2. See error

**Expected behavior**
Database support to longer room ids and import does not fail.

PR #144, opened 2021-05-31
Extend room id to support new greenlight room ids

dennis95stumm:
# Fixes \#143

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Extended room id to support ids of new greenlight versions (e.g. xxx-xxx-xxx-xxx)

### DIFF
--- a/database/migrations/2021_05_31_164031_extend_room_id.php
+++ b/database/migrations/2021_05_31_164031_extend_room_id.php
@@ -34,6 +34,7 @@ class ExtendRoomId extends Migration
      */
     public function down()
     {
+        Schema::disableForeignKeyConstraints();
         Schema::table('rooms', function (Blueprint $table) {
             $table->string('id',11)->change();
         });
@@ -46,5 +47,6 @@ class ExtendRoomId extends Migration
         Schema::table('room_user', function (Blueprint $table) {
             $table->string('room_id',11)->change();
         });
+        Schema::enableForeignKeyConstraints();
     }
 }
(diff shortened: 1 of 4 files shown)


database/migrations/2021_05_31_164031_extend_room_id.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('rooms', function (Blueprint $table) {
```
SamuelWei, line 37: The rollback is throwing the error 

`Cannot change column 'id': used in a foreign key constraint 'meetings_room_id_foreign' of table 'pilos.meetings'`

dennis95stumm: The question is, whether a rollback is possible or not? I can fix this bug but in case if there are rooms with longer ids in the database (e.g. imported from greenlight) they will be cutted and then it wouldn't work!

dennis95stumm: @SamuelWei see commit 64c845c. But as I said, this migration can't be rolled back if there are longer ids!

SamuelWei: @dennis95stumm I'm fine with this solution, just wanted to make sure the immediate rollback after a migration is possible

SamuelWei (approved):

Issue #72, opened 2020-11-02
Personalised room links


SamuelWei:
**Is your feature request related to a problem? Please describe.**
It should be possible to create a personalised room link for a meeting. This is for those cases where the general allow guest rule it to broad, but it would also make no sence to create a new user account for this person.

**Describe the solution you'd like**
Add a new section to the room overview, next to members and files. or implement feature into the members section.

Clicking on a create button opens a modal, asking for the persons name and role (participant or moderator). Once completed, a link should be shown. All personalised room links should be shown in a table, so the name and role can be edited and removed. The actual link could be hidden behind a button to make the table more clean.

If someone is using this link, they should not be required to enter a room acccess code. They can't modify their name, as it is set by the data in the database.

**UI ideas**
![persroomlinks](https://user-images.githubusercontent.com/4281791/120449664-ef739080-c38f-11eb-8e24-d6358a516a37.PNG)
![image](https://user-images.githubusercontent.com/4281791/120450494-c7d0f800-c390-11eb-870c-b8bf17fe2086.png)

dennis95stumm:
@SamuelWei Should the room access code really be not necessary? For logged in users the access code is also necssray!

PR #145, opened 2021-06-04
Personalized room links

dennis95stumm:
# Fixes \#72

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist
- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Added personalized room links


resources/js/components/Room/TokensComponent.vue
```diff
@@ -0,0 +1,397 @@
...
+      bvModalEvt.preventDefault();
+
+      this.actionRunning = true;
+
+      const config = {
+        method: this.model.token == null ? 'post' : 'put',
+        data: _.cloneDeep(this.model)
+      };
+
+      Base.call(`rooms/${this.room.id}/tokens/${this.model.token == null ? '' : this.model.token}`, config).then(response => {
```
SamuelWei, line 292: if this.model.token == null the url is `rooms/ROOM-ID/tokens/ `and not `rooms/ROOM-ID/tokens` The additional tailing slash results in a redirect and fails as a result

That might work
` Base.call(`rooms/${this.room.id}/tokens${this.model.token == null ? '' : '/'+this.model.token}`, config)`

dennis95stumm: I cannot reproduce that!

SamuelWei: ok, I'll have another look once the PR is done. It may just have been an issue with my local setup.

Issue #1289, opened 2024-08-12
Delete Room Member Dialog not closed on 410 error


Sabr1n4W:
**Describe the bug**
When a 410 error (user is not a member of the room (anymore)) occurs while deleting a room member the dialog to delete the room member stays visible. The name within the dialog is changed to the name of the next user in the member list and the dialog can be used to delete this room member.

**To Reproduce**
1. Open a room that has more than 1 member
2. Open the delete dialog for the first room member and delete the room member in another tab
3. Now try to delete the room member through the open delete dialog (should lead to 410 error because the user is not am member of the room anymore)
4. The room member is deleted from the members list but the delete dialog stays visible and contains the name of the next user inside the room member list

**Expected behavior**
The delete room member dialog should be closed when a 410 error occurs.

**Additional context**
If the dialog remains open and only the user's name is changed, this could lead to other members being deleted without the user deleting the member noticing.

PR #1290, opened 2024-08-12
Fix delete room member dialog not closed on 410 error

Sabr1n4W:
# Fixes \#1289

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist
- [ ] Code updated to current develop branch head
- [ ] Passes CI checks
- [x] Is a part of an issue
- [ ] Changelog is updated

## Changes

- fix delete room member dialog not closed on 410 error

### DIFF
--- a/resources/js/components/RoomTabMembersDeleteButton.vue
+++ b/resources/js/components/RoomTabMembersDeleteButton.vue
@@ -85,6 +85,7 @@ function deleteMember () {
     if (error.response) {
       // user not found
       if (error.response.status === env.HTTP_GONE) {
+        showModal.value = false;
         emit('deleted');
         return;
       }


SamuelWei (changes requested):
I would only remove the return, therefore the error message is shown and the modal is closed

Sabr1n4W:
> I would only remove the return, therefore the error message is shown and the modal is closed

In the RoomTabMembersEdit button the reaction to a 410 error seems to be the same to the code I suggested for the RoomTabMembersDeleteButton.  

Should I still change the code inside the RoomTabMembersDeleteButton so that only the return is removed?

Sabr1n4W:
> I would only remove the return, therefore the error message is shown and the modal is closed

In the RoomTabMembersEditButton the reaction to a 410 error seems to be the same to the code I suggested for the RoomTabMembersDeleteButton.  

Maybe the code should also be changed there so that the error handling is similar.

SamuelWei:
> > I would only remove the return, therefore the error message is shown and the modal is closed
> 
> In the RoomTabMembersEditButton the reaction to a 410 error seems to be the same to the code I suggested for the RoomTabMembersDeleteButton.
> 
> Maybe the code should also be changed there so that the error handling is similar.

Agree, could you please include this change in this PR

SamuelWei (approved):

PR #1548, opened 2024-11-12
RoomTypeChangeButton fix error and loading state

Sabr1n4W:
# Fixes \#

-   **Bugfix**
-   Feature
-   Documentation
-   Refactoring (e.g. Style updates, Test implementation, etc.)
-   Other (please describe):

## Checklist

-   [x] Code updated to current develop branch head
-   [x] Passes CI checks
-   [ ] Is a part of an issue
-   [ ] Tests added for the bugfix or newly implemented feature, describe below why if not
-   [x] Changelog is updated
-   [ ] Documentation of code and features exists

## Changes

- Fix error and loading state of the RoomTypeChangeButton dialog


resources/js/components/RoomTypeChangeButton.vue
```diff
@@ -101,6 +118,11 @@ function handleOk() {
   }
 }
 
+function resetRoomTypeSelectEmits() {
+  roomTypeSelectBusy.value = false;
+  roomTypeSelectLoadingError.value = false;
+}
+
```
samuelwei: What are the pros/cons for adding this to the `editRoomType` function

Sabr1n4W: The main difference is that the current solution resets the values when closing the dialog and that a solution with adding this to the `editRoomType` function would reset the values when opening the dialog. 

Pros for adding this to the `editRoomType` function:
* No new function just for resetting when we already have a function where the dialog data is reset
* Resetting is handled in a way that is similar to how we normally handle resetting the dialogs and `@hide` inside the dialog would not be needed anymore _(only other place where `@hide` is used is inside the RoomCreateComponent)_

Cons:
- Reset is currently handled similar to the RoomCreateComponent which also contains the RoomTypeSelect and therefore `roomTypeSelectBusy` and `roomTypeSelectLoadingError`
- Current function name is a little clearer (clearer that values are reset)


I think adding this to the `editRoomType` function would be the better solution but I think that we then should perhaps also think about modifying the RoomCreateComponent to something similar

samuelwei: Agree, the handling should be consistent

samuelwei (approved):

PR #1801, opened 2025-01-22
Fix missing form validation feedback on password fields

samuelwei:
## Type

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist

- [x] Passes CI checks
- [ ] Is a part of an issue
- [x] Changelog is updated

## Changes

- Missing form validation feedback for password fields on login page

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved form validation feedback for password fields on login page.

- **Tests**
  - Enhanced e2e tests for login error handling.
  - Added checks for `aria-invalid` attributes on input fields during error scenarios.

- **Documentation**
  - Updated CHANGELOG.md to document form validation improvements.


samuelwei:
@coderabbitai full review

tests/Frontend/e2e/Login.cy.js
Sabr1n4W, line 1: Consider using the id of the input fields to check for the invalid state and using `.should("have.attr", ...)`

e.g. instead of

```js
cy.get('[data-test="email-field"] input')
      .invoke("attr", "aria-invalid")
      .should("eq", "true");
``` 
think about using something like
```js
cy.get("#local-email").should("have.attr", "aria-invalid", "true");
``` 

This would also better fit the way you currently use to check if the attribute is not set and the way we normally check attribute values in other tests.

Sabr1n4W, line 1: Consider using the id of the input fields to check for the invalid state and using `.should("have.attr", ...)`

e.g. instead of

```js
cy.get('[data-test="email-field"] input')
      .invoke("attr", "aria-invalid")
      .should("eq", "true");
``` 
think about using something like
```js
cy.get("#local-email").should("have.attr", "aria-invalid", "true");
``` 

This would also better fit the way you currently use to check if the attribute is not set and the way we normally check attribute values in other tests.

Sabr1n4W, line 1: Checks that the invalid state of the password input are reset is still missing.

Could be added after checking that the errror messages are hidden again (After the 429 error)

Sabr1n4W (changes requested):

tests/Frontend/e2e/Login.cy.js
```diff
@@ -410,6 +419,9 @@ describe("Login", function () {
       'app.flash.server_error.error_code_{"statusCode":500}',
     ]);
 
+    // Check password field invalid state is removed
+    cy.get("#local-password").should("not.have.attr", "aria-invalid");
+
```
Sabr1n4W: Maybe move this before the 500 error after checking that the 422 error messages are reset after the 429 error (lines 398 to 400)

Sabr1n4W (changes requested):

Sabr1n4W (approved):

PR #2015, opened 2025-04-03
Fix action column

samuelwei:
## Type

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist

- [x] Passes CI checks
- [ ] Is a part of an issue
- [x] Changelog is updated

## Changes

- Fix width for action column with only one item
- Fix wrong calculation how many actions are visible
- Fix action column not always shown in user list

## Other information
- If only one item should be shown the column width has been set to 5rem, however that is too small for some locales to display the column heading text -> increases to 8rem.
- The previous code to determine the number of visible action buttons has a bug where the return inside the foreach loop to check if a required permission is not met doesn't exit the parent filter action with false as indentend, but only the foreach loop. The code has been changed so that the return correctly exits the filter function.
- The action column should always be shown in the users list, as the view and edit action are always shown for the current user. As we determining if the current user is on the current page is a lot more complex, this column is always visible

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved issues with the admin UI action column sizing in different locales.
  - Fixed the display so that the action column is hidden when there are no permitted actions.
  - Addressed visibility issues with the action column in the user list.

- **Styles**
  - Increased the width of the action column on larger screens for enhanced readability.

- **Tests**
  - Updated assertions in tests to reflect changes in the UI structure and permissions model for admin roles, room types, server pools, and servers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow removing English from the list of enabled locales ([#1919], [#1920])
 - Action column in admin UI too small with only one action in some locales ([#2015])
 - Action column in admin UI is shown even if no action is available due to permissions ([#2015])
+- Action column in admin UI user list no always shown ([#2015])
 
 ## [v4.3.1] - 2025-02-17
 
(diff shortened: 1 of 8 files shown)


samuelwei:
Frontend tests all pass locally

PR #2116, opened 2025-05-05
Static dialin pin

q16marvin:
# Fixes \#

## Type

- Bugfix
- **Feature**
- Documentation
- Other (please describe):

## Checklist

- [ ] Passes CI checks
- [ ] Is a part of an issue
- [ ] Changelog is updated

## Changes

We have everyday on small meeting, if someone is on the road he must dialin via phone. Currently BBB create on every new Meeting a new random dialin pin. So for specified Meetings we need a fixed PIN that never changed.

This Change bringt a new Room Field:

grafik

## Summary by CodeRabbit

- **New Features**
  - Introduced dial-in PIN functionality for rooms, allowing generation, assignment, and validation of a unique 5-digit PIN.
  - Added support for dial-in PIN settings at both room and room type levels, including enforcement and default options.
  - New UI components and form fields for managing dial-in PINs in room and admin settings.
  - Dial-in PIN is now included in meeting creation parameters and room settings displays.

- **Localization**
  - Added English and German translations for all new dial-in PIN related labels, messages, and validation attributes.

- **Bug Fixes**
  - Ensured proper validation and uniqueness of dial-in PINs when creating or updating rooms.


samuelwei:
Hello @q16marvin,

Thank you for your PR. Unfortunately, adding static dial-in PINs is not that easy to accomplish.
In order to have a predictable combination of dial-in number and PIN, we need to know not only the PIN, but also the number.
Since most users have not just one, but several BBB servers, they all use the same Freeswitch server with a common number.
All BBB servers must then use a central dial-in plan to resolve the pin for a meeting.

Scalelite can provide such a dial-in plan, but in PILOS you can also add the servers directly, so PILOS would also need to provide a dial-in plan. The combination of other load balancers and adding servers directly makes the whole thing even more complex, the dial in plans would have to be merged by something.

In the event that other frontends can also create meetings on the same BBB server, we also need to ensure that the same number is not assigned twice, which is particularly difficult as BBB creates these numbers when they are not explicitly set. So it can happen that a meeting without a fixed PIN is assigned the same PIN by BBB as a room with a static PIN.

If I am wrong somewhere, please correct me. It's been a few days since I last dealt with this complex issue.

q16marvin:
Hi @SamuelWei,

you are absolutly right, in big enviroments with some BBB Servers and maybe other frontends to manage this BBB Servers it can happen colliding PINs :(

we have only one BBB Server and with a freeswitch dial in. Could we add this function with a .env setting called like "PILOS_DIALIN_PIN_ENABLE = true" and this stuff is only useable if this is true and document it with the "risky in big enviroments"?

Or other Idea: BBB Servers generate Random PINs from 10000 - 99999. Pilos could use Pins from 00001 - 09999, than there would never be a conflict.

thats the last "feature" why we cant move from our self developed stupid frontend to your extrem cool frontend for BBB. So i would be realy happy if we find any compromise :)

PR #2282, opened 2025-07-10
Fix: Infinite loading when navigating back after logout redirect due to bfcache

samuelwei:
## Type

- **Bugfix**
- Feature
- Documentation
- Other (please describe):

## Checklist

- [x] Passes CI checks
- [ ] Is a part of an issue
- [x] Changelog is updated

## Changes

- Fixed infinite loading when navigating back after logout redirect due to bfcache

## Other information
Safari and Chrome are creating a bfcache , ignoring the no-cache header.
Firefox currently respects the no-cache header and does not create a bfcache.

Without this PR the JS and DOM state is restored after navigating back from a logout redirect. During logout the page is covered with a loading spinner. If the user returns to this state, the loading never finishes. This PR adds a new listener to handle the restoration from the bfcache, redirects the user to the logout page and closes the loading spinner overlay.

I'm unable to write a cypress test that triggers the bfcache, therefore this code cannot be tested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where navigating back after logout could result in an infinite loading state due to browser cache restoration. Users are now properly redirected and the loading spinner is cleared in this scenario.

* **Documentation**
  * Updated the changelog to reflect the fix for the infinite loading issue after logout.

### DIFF
--- a/resources/js/components/MainNav.vue
+++ b/resources/js/components/MainNav.vue
@@ -329,6 +329,19 @@ const loginRoute = computed(() => {
   return loginRoute;
 });
 
+/**
+ * Handle the page is restored from the back/forward cache after logout.
+ * The loading spinner is still visible, so we need to remove it
+ * and redirect the user to the logout page.
+ */
+async function pageShownAfterLogoutHandler(event) {
+  window.removeEventListener("pageshow", pageShownAfterLogoutHandler);
+  if (event.persisted) {
+    await router.push({ name: "logout" });
+    loadingStore.setLoadingFinished();
+  }
+}
+
 async function logout() {
   let response;
   try {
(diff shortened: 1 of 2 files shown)


resources/js/components/MainNav.vue
```diff
@@ -342,6 +355,9 @@ async function logout() {
   }
 
   if (response.data.redirect) {
+    // Add listener for when user returns to this page
+    // without a full page reload, state is restored from back/forward cache
```
danielmachill, line 359: // Listen for page restores from browser bfcache to ensure logout state is enforced

danielmachill (approved):

PR #2477, opened 2025-09-26
Migrate to Tailwind v4

samuelwei:
## Type

- Bugfix
- Feature
- Documentation
- Other (please describe):

## Checklist

- [x] Passes CI checks
- [ ] Is a part of an issue
- [x] Changelog is updated

## Changes

- Upgrade to Tailwind v4

## Release Notes

* **New Features**
  * Added dark mode support across the application with enhanced theming capabilities.

* **Updates**
  * Upgraded Tailwind CSS to v4 with improved styling architecture.
  * Updated Node.js minimum requirement to version 20.
  * Migrated styling infrastructure for better performance and maintainability.

* **Bug Fixes**
  * Fixed badge component height and padding inconsistencies.

* **Style**
  * Refined spacing and visual alignment across UI components for consistency.


samuelwei:
TODO: Remove SASS npm package, check if postcss is still needed

Sabr1n4W:
> TODO: Remove SASS npm package, check if postcss is still needed

Maybe it should also be checked if autoprefixer could be removed.

samuelwei:
> > TODO: Remove SASS npm package, check if postcss is still needed
> 
> Maybe it should also be checked if autoprefixer could be removed.

Done; all three can be removed

Sabr1n4W (changes requested):
**Not completely done reviewing this** 

but found an issue with the RoomCardSkeleton:

**Previous:**
Screenshot 2025-10-06 145358


**Current:**
Screenshot 2025-10-06 145336

Seems to be caused by the usage of `h-100` inside the component (`RoomCardSkeleton.vue`)

<hr>

Fixed in e26bf7a69db85eb1e6eb164fa2cfee684f463a1b

Sabr1n4W (changes requested):
**Still not completely done reviewing this** but

Other difference found in the help dialog of the AdminRolesView:

Previous:
Screenshot 2025-10-06 161418


Current:
Screenshot 2025-10-06 161414


Main difference: Table now uses the full width


Personally I don't think that it is bad that the table now uses the full width but think that the centered headers maybe make it a bit harder to read the table when the table uses the full width.

This could maybe be fixed with a bit of space above the headers / table or by moving the headers to the left.


<hr>

Fixed in 0aa3f1b009b948791b8c06a8aff071c2e28e48c1, 6ea1ca8b4192971700cc5d8b96e0e618a049204e

Sabr1n4W (changes requested):
**Still not completely done reviewing this** but

Mouse feedback for ColorSelect seems to be missing after changes in this pull request

Previous:
previous mouse feedback



Current:
ezgif-332bdf41eb9b0c



Behaviour was tested for the browsers Firefox and Chrome

<hr>

Fixed in 464494fcb93c07b917ac92a2f828ca6bb8d29dff, 421c248156b1fd16837c82b1c647e6ec2ed8c14c

Sabr1n4W (commented):
**Still not completely done reviewing this** but 

found some additional differences concerning borders and dividers when the dark mode is activated. 

In my opinion these are mostly improvements that don't need to be corrected. 

### Previous:
* Room Tab Files:
Room Tabs files difference previous


* Room Tab Personalized links:
Room Tabs pers links difference previous


* Settings:
Settings images diff prev
Settings bbb diff prev



### Current:
* Room Tab Files:
Room Tabs files difference current


* Room Tab Personalized links:
Room Tabs pers links difference current


* Settings:
Settings images diff current
Settings bbb diff current



(The only room tabs that previously had another color are the room tab files and room tab personalized links that are shown in the pictures. All other tabs did not change in color and already looked similar to the tabs shown here in the current version)

samuelwei:
> **Still not completely done reviewing this** but
> 
> found some additional differences concerning borders and dividers when the dark mode is activated.
> 
> In my opinion these are mostly improvements that don't need to be corrected.
> 
> ### Previous:
> * Room Tab Files:
> 
> Room Tabs files difference previous
> * Room Tab Personalized links:
> 
> Room Tabs pers links difference previous
> * Settings:
> 
> Settings images diff prev Settings bbb diff prev
> ### Current:
> * Room Tab Files:
> 
> Room Tabs files difference current
> * Room Tab Personalized links:
> 
> Room Tabs pers links difference current
> * Settings:
> 
> Settings images diff current Settings bbb diff current
> (The only room tabs that previously had another color are the room tab files and room tab personalized links that are shown in the pictures. All other tabs did not change in color and already looked similar to the tabs shown here in the current version)

I noticed that after the Tailwind migration, the border colours were too bright in dark mode, in my opinion, so I changed them.

I thought the colour change was caused by the Tailwind migration, but apparently it was already like that before the migration.

I agree – the more subtle border colours are better and should be kept.

Sabr1n4W (commented):
Done reviewing this and could not find any more differences between the current an previous version.


I am currently not completely sure about the change of the table header "_Authentication Type_" to "_Auth. Type_". I think that it is a lot better when resizing the window and for different window sizes but personally think that it seems a bit inconsistent because we mostly don't use shortened labels everywhere else.

samuelwei:
> Done reviewing this and could not find any more differences between the current an previous version.
> 
> I am currently not completely sure about the change of the table header "_Authentication Type_" to "_Auth. Type_". I think that it is a lot better when resizing the window and for different window sizes but personally think that it seems a bit inconsistent because we mostly don't use shortened labels everywhere else.

Change "_Authentication Type_" to "Login method" (EN), keep german term, "_Type d'identification_" "_Méthode de connexion_" (FR)

Add custom css
```css
.p-datatable-header-cell {
overflow-wrap: anywhere;
hyphens: auto;
}
```

Sabr1n4W:
> **Still not completely done reviewing this** but
> 
> Mouse feedback for ColorSelect seems to be missing after changes in this pull request
> 
> Previous: previous mouse feedback 
> 
> Current: ezgif-332bdf41eb9b0c
> 
> Behaviour was tested for the browsers Firefox and Chrome

I found an additional problem similar to this that already existed before.

For the **disabled state** of the color select the classes `pointer-events-none` and `cursor-not-allowed` are set but the class  `pointer-events-none`  seems to prevent the effects of `cursor-not-allowed` (changed curser that indicates that clicking is not possible is not shown because of `pointer-events-none`). 


https://github.com/user-attachments/assets/39b9a161-5781-4710-81ef-84f2115deb20



Therefore the class `cursor-not-allowed` could either be removed completely or another way to add this class to apply the cursor change would have to be found.

Personally i think that changing the cursor in the disabled state definitely makes the disabled state a lot clearer but is not absolutely necessary. Therefore i think that removing the class `cursor-not-allowed` would be an option.

Sabr1n4W:
> **Still not completely done reviewing this** but
> 
> Other difference found in the help dialog of the AdminRolesView:
> 
> Previous: Screenshot 2025-10-06 161418
> 
> Current: Screenshot 2025-10-06 161414
> 
> Main difference: Table now uses the full width
> 
> Personally I don't think that it is bad that the table now uses the full width but think that the centered headers maybe make it a bit harder to read the table when the table uses the full width.
> 
> This could maybe be fixed with a bit of space above the headers / table or by moving the headers to the left.

Table with headers moved to the left:
Screenshot 2025-10-08 174730

Table with headers moved to the left and fixed padding:
Screenshot 2025-10-08 183151

Sabr1n4W:
> > Done reviewing this and could not find any more differences between the current an previous version.
> > I am currently not completely sure about the change of the table header "_Authentication Type_" to "_Auth. Type_". I think that it is a lot better when resizing the window and for different window sizes but personally think that it seems a bit inconsistent because we mostly don't use shortened labels everywhere else.
> 
> Change "_Authentication Type_" to "Login method" (EN), keep german term, "_Type d'identification_" "_Méthode de connexion_" (FR)
> 
> Add custom css
> 
> ```css
> .p-datatable-header-cell {
> overflow-wrap: anywhere;
> hyphens: auto;
> }
> ```

Adding the custom css does not work really well. 
E.g. the `overflow-wrap: anywhere` would result in something like the following when the window is relatively small (user already has to scroll to see the full table):

Screenshot 2025-10-09 164841

with `overflow-wrap: break-word` this could be improved a bit but would still look something like the following in german when the window is relatively small (user already has to scroll to see the full table):

Screenshot 2025-10-09 165051

This seems a bit unnecessary to me when the user already has to scroll to see the full table and I personally would prefer the headers without word breaks in this context.

With bigger window sizes the results are a bit better but sometimes still not really well (especially in german) because in my opinion the word breaks sometimes make it a bit harder to read the headers. 

e.g.
Screenshot 2025-10-09 170814

Because of all this I would prefer to not add the cutom css.

One advantage of adding the word breaks would be that the table could sometimes be fully displayed, when previously the user had to scroll to see the entire table.

samuelwei:
> > > Done reviewing this and could not find any more differences between the current an previous version.
> > > I am currently not completely sure about the change of the table header "_Authentication Type_" to "_Auth. Type_". I think that it is a lot better when resizing the window and for different window sizes but personally think that it seems a bit inconsistent because we mostly don't use shortened labels everywhere else.
> > 
> > 
> > Change "_Authentication Type_" to "Login method" (EN), keep german term, "_Type d'identification_" "_Méthode de connexion_" (FR)
> > Add custom css
> > ```css
> > .p-datatable-header-cell {
> > overflow-wrap: anywhere;
> > hyphens: auto;
> > }
> > ```
> 
> Adding the custom css does not work really well. E.g. the `overflow-wrap: anywhere` would result in something like the following when the window is relatively small (user already has to scroll to see the full table):
> 
> Screenshot 2025-10-09 164841
> with `overflow-wrap: break-word` this could be improved a bit but would still look something like the following in german when the window is relatively small (user already has to scroll to see the full table):
> 
> Screenshot 2025-10-09 165051
> This seems a bit unnecessary to me when the user already has to scroll to see the full table and I personally would prefer the headers without word breaks in this context.
> 
> With bigger window sizes the results are a bit better but sometimes still not really well (especially in german) because in my opinion the word breaks sometimes make it a bit harder to read the headers.
> 
> e.g. Screenshot 2025-10-09 170814
> 
> Because of all this I would prefer to not add the cutom css.
> 
> One advantage of adding the word breaks would be that the table could sometimes be fully displayed, when previously the user had to scroll to see the entire table.

maybe adding it depending on the tables layout style could help? See resources/css/app/_action-column.css

Sabr1n4W:
> > > > Done reviewing this and could not find any more differences between the current an previous version.
> > > > I am currently not completely sure about the change of the table header "_Authentication Type_" to "_Auth. Type_". I think that it is a lot better when resizing the window and for different window sizes but personally think that it seems a bit inconsistent because we mostly don't use shortened labels everywhere else.
> > > 
> > > 
> > > Change "_Authentication Type_" to "Login method" (EN), keep german term, "_Type d'identification_" "_Méthode de connexion_" (FR)
> > > Add custom css
> > > ```css
> > > .p-datatable-header-cell {
> > > overflow-wrap: anywhere;
> > > hyphens: auto;
> > > }
> > > ```
> > 
> > 
> > Adding the custom css does not work really well. E.g. the `overflow-wrap: anywhere` would result in something like the following when the window is relatively small (user already has to scroll to see the full table):
> > Screenshot 2025-10-09 164841
> > with `overflow-wrap: break-word` this could be improved a bit but would still look something like the following in german when the window is relatively small (user already has to scroll to see the full table):
> > Screenshot 2025-10-09 165051
> > This seems a bit unnecessary to me when the user already has to scroll to see the full table and I personally would prefer the headers without word breaks in this context.
> > With bigger window sizes the results are a bit better but sometimes still not really well (especially in german) because in my opinion the word breaks sometimes make it a bit harder to read the headers.
> > e.g. Screenshot 2025-10-09 170814
> > Because of all this I would prefer to not add the cutom css.
> > One advantage of adding the word breaks would be that the table could sometimes be fully displayed, when previously the user had to scroll to see the entire table.
> 
> maybe adding it depending on the tables layout style could help? See resources/css/app/_action-column.css

Works a lot better but still a bit tricky.

`overflow-wrap: anywhere` results in everything beeing broken when necessary, but also results in words being broken without a hyphen added (e.g. english words that are capitalized)
`overflow-warp: break-word` only breaks words where a hyphen can be inserted (because of this long english words that are capitalized will for example not be broken)

Correction:
`overflow-warp: break-word` does not seem to have any effect at all here, but `hyphens: auto` seems to be responsible for the break when this is set

So the options seem to be 
* Only break words when a hyphen can be added
* Always break which would (especially in english) result in words being broken without the usage of hyphens

Results for `overflow-warp: break-word` / only `hyphens: auto` (smallest window size where the table is still fully visible):
Screenshot 2025-10-10 071149
Screenshot 2025-10-10 071224

Results for `overflow-wrap: anywhere` (smallest window size where the table is still fully visible):
Screenshot 2025-10-10 071500

Screenshot 2025-10-10 071517
(Picture header will also be broken on bigger window sizes)

Sabr1n4W:
> > > > > Done reviewing this and could not find any more differences between the current an previous version.
> > > > > I am currently not completely sure about the change of the table header "_Authentication Type_" to "_Auth. Type_". I think that it is a lot better when resizing the window and for different window sizes but personally think that it seems a bit inconsistent because we mostly don't use shortened labels everywhere else.
> > > > 
> > > > 
> > > > Change "_Authentication Type_" to "Login method" (EN), keep german term, "_Type d'identification_" "_Méthode de connexion_" (FR)
> > > > Add custom css
> > > > ```css
> > > > .p-datatable-header-cell {
> > > > overflow-wrap: anywhere;
> > > > hyphens: auto;
> > > > }
> > > > ```
> > > 
> > > 
> > > Adding the custom css does not work really well. E.g. the `overflow-wrap: anywhere` would result in something like the following when the window is relatively small (user already has to scroll to see the full table):
> > > Screenshot 2025-10-09 164841
> > > with `overflow-wrap: break-word` this could be improved a bit but would still look something like the following in german when the window is relatively small (user already has to scroll to see the full table):
> > > Screenshot 2025-10-09 165051
> > > This seems a bit unnecessary to me when the user already has to scroll to see the full table and I personally would prefer the headers without word breaks in this context.
> > > With bigger window sizes the results are a bit better but sometimes still not really well (especially in german) because in my opinion the word breaks sometimes make it a bit harder to read the headers.
> > > e.g. Screenshot 2025-10-09 170814
> > > Because of all this I would prefer to not add the cutom css.
> > > One advantage of adding the word breaks would be that the table could sometimes be fully displayed, when previously the user had to scroll to see the entire table.
> > 
> > 
> > maybe adding it depending on the tables layout style could help? See resources/css/app/_action-column.css
> 
> Works a lot better but still a bit tricky.
> 
> `overflow-wrap: anywhere` results in everything beeing broken when necessary, but also results in words being broken without a hyphen added (e.g. english words that are capitalized) `overflow-warp: break-word` only breaks words where a hyphen can be inserted (because of this long english words that are capitalized will for example not be broken)
> 
> Correction: `overflow-warp: break-word` does not seem to have any effect at all here, but `hyphens: auto` seems to be responsible for the break when this is set
> 
> So the options seem to be
> 
>     * Only break words when a hyphen can be added
> 
>     * Always break which would (especially in english) result in words being broken without the usage of hyphens
> 
> 
> Results for `overflow-warp: break-word` / only `hyphens: auto` (smallest window size where the table is still fully visible): Screenshot 2025-10-10 071149 Screenshot 2025-10-10 071224
> 
> Results for `overflow-wrap: anywhere` (smallest window size where the table is still fully visible): Screenshot 2025-10-10 071500
> Screenshot 2025-10-10 071517 (Picture header will also be broken on bigger window sizes)

Decided against adding custom css.
Could be useful for some cases but adds several problems and increases complexity when applied in general.

Changes decided for _Authentication Type_, _Authenticator_ and _Authenticator-ID_:

| Language |  Authentication Type / Authenticator | Authenticator-ID |
| -- | -- | -- | 
| English | Login method | External user ID |
| German | Login-Methode | Externe Benutzer-ID |
| French | Méthode de connexion | _Not decided yet_ |

<hr>

Moved to #2551

PR #2726, opened 2026-01-14
Refactor file access

### LINKED ISSUE #1409: Refactor file access
Currently, room files are listed and when clicked, an async function is triggered to generate a short live link to download the file.
This is a problem as some browsers block the opening of a new window that is not directly caused by a user action.

Currently, the room token and room access code are sent in the header of all API requests to authenticate a user against the room.
A file access link is not an API request, so setting the header is not possible.

A new approach could be to generate an encrypted auth token when a user authenticates against the room via the api and associate this with the user's session. This token could be sent via a get request parameter during file access to check if the access code or token is still valid, while still protecting the token/code from being exposed in server logs, etc.

Sabr1n4W:
# Implements \#1409

## Type

- Bugfix
- Feature
- Documentation
- Other (please describe):

## Checklist

- [x] Passes CI checks
- [X] Is a part of an issue
- [x] Changelog is updated

## Changes

- Added room auth request that exchanges the provided access code or token for a room auth token 
  -  The room auth token is used in further API requests to authenticate the user against the room
- Removed the generation and usage of short live links to download files
  -  Instead the room auth token is passed in the file access link

## Summary by CodeRabbit

* **New Features**
  * Room authentication now supports session-scoped auth tokens and personalized room links.
  * File access now uses signed URLs and a terms-of-use flow for safer downloads.

* **Bug Fixes**
  * Fixed file downloads being blocked in strict-browser scenarios.

* **UI / Documentation**
  * “Tokens” renamed to “Personalized links” across the UI and translations.
  * Improved error pages and clearer authentication/error messages; settings labels updated.


resources/js/components/RoomTabFilesViewButton.vue
```diff
@@ -1,14 +1,29 @@
...
     :disabled="disabled"
     target="_blank"
+    rel="opener"
+    :href="downloadUrl"
     :loading="loading"
     icon="fa-solid fa-eye"
     data-test="room-files-view-button"
-    @click="downloadFile"
+    as="a"
   />
```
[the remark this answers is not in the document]
Sabr1n4W: `rel="opener"` needed here to make error handling possible

resources/js/views/RoomsView.vue
```diff
@@ -295,7 +302,18 @@ onMounted(() => {
...
+    authenticate(ROOM_AUTH_TOKEN_TYPE_PERSONALIZED_LINK, props.token).then(
+      (success) => {
+        if (success) {
+          load();
+        }
+      },
+    );
+  } else {
+    load();
+  }
```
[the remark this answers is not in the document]
Sabr1n4W: Was temporary fixed but most likely needs improvements / more testing

tests/Frontend/e2e/RoomsViewFilesFileActions.cy.js
```diff
@@ -1258,5 +1081,60 @@ describe("Rooms view files file actions", function () {
...
+      $window.postMessage(message, Cypress.config("baseUrl"));
+    });
+
+    // Check that files are still there and toast message is not shown
+    cy.get('[data-test="room-file-item"]').should("have.length", 3);
+
+    cy.get(".p-toast-message").should("not.exist");
+
+    // Check that files were not reloaded
+    cy.get("@reloadRoomFilesRequest").should("be.null");
```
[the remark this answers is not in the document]
Sabr1n4W: `@alias.all` is still an undocumented feature

resources/js/views/RoomsView.vue
```diff
@@ -548,10 +590,112 @@ function setPageTitle(roomName) {
...
+    api
+      .call(url, {
+        method: "POST",
+        data: data,
+      })
+      .then((response) => {
+        if (response.status !== 204) {
+          // Set room auth token for further requests if response is not empty
+          roomAuthToken.value = response.data.data;
+        }
```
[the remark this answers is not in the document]
Sabr1n4W: `authenticate` should only be called if no roomAuthToken is present (first authenticate request or authenticate after roomAuthToken was reset after an error (`handleInvalidRoomAuthToken`)). Therefore resetting the roomAuthToken should not be necessary here

app/Http/Middleware/RoomAuthenticate.php
```diff
@@ -37,25 +40,84 @@ public function handle($request, Closure $next, $allowUnAuthenticated = false)
...
+
+                return $this->handleError(CustomErrorMessages::ROOM_INVALID_TOKEN->value, 401, 'Invalid token', __('rooms.flash.auth_token_invalid'));
             }
 
-            $token->last_usage = now();
-            $token->save();
+            // Retrieve the validated input
+            $validated = $validator->validated();
+            $providedRoomAuthToken = $validated['room_auth_token'];
+            $providedRoomAuthTokenType = RoomAuthTokenType::from($validated['room_auth_token_type']);
```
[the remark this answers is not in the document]
Sabr1n4W: Currently seems to work. ::from seems to also accept and cast string arguments, but not completely sure about that.

tests/Backend/Feature/api/v1/Room/FileTest.php
```diff
@@ -146,74 +153,125 @@ public function test_view_files()
...
+            'type' => RoomAuthTokenType::CODE,
+        ]);
+
+        // Testing guests with room auth token (access code)
+        $this
+            ->getJson(route('api.v1.rooms.files.get', [
+                'room' => $this->room,
+                'room_auth_token' => $roomAuthToken->id,
+                'room_auth_token_type' => RoomAuthTokenType::CODE->value,
+            ]))
```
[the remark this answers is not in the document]
Sabr1n4W: Room auth token table has an id column.

room_tokens table was a different table that was renamed to room_personalized_links

app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -423,4 +431,99 @@ public function transferOwnership(Room $room, TransferOwnershipRequest $request)
...
+                Counter::get('room_authentication_errors_total')->inc('access_code_invalid');
+                Log::notice('Room access code authentication failed for room {room}', ['room' => $room->getLogLabel()]);
+
+                // Increment rate limit counter for failed access code attempts
+                RateLimiter::increment($rateLimitKey);
+
+                abort(401, CustomErrorMessages::ROOM_INVALID_CODE->value);
+            }
+        } elseif ($request->type === RoomAuthTokenType::PERSONALIZED_LINK->value) {
+            if (! Auth::guest()) {
```
samuelwei:
```suggestion
            if (Auth::user()) {
```

app/Http/Controllers/api/v1/RoomController.php
```diff
@@ -423,4 +431,99 @@ public function transferOwnership(Room $room, TransferOwnershipRequest $request)
...
+     */
+    public function authenticate(Room $room, RoomAuthRequest $request)
+    {
+        // Check if user tries to authenticate even though it is not necessary
+        if (\Illuminate\Support\Facades\Auth::user() && ($room->owner->is(Auth::user()) || $room->members->contains(Auth::user()) || Auth::user()->can('viewAll', Room::class))) {
+            return response()->noContent();
+        }
+
+        if ($request->type === RoomAuthTokenType::CODE->value) {
+            if (! $room->getRoomSetting('allow_guests') && ! Auth::user()) {
```
samuelwei:
```suggestion
            if (! $room->getRoomSetting('allow_guests') && Auth::guest()) {
```

resources/js/components/RoomTabFiles.vue
```diff
@@ -442,8 +437,43 @@ function onPage(event) {
...
+  } else if (event.data.type === HTTP_ROOM_INVALID_AUTH_TOKEN) {
+    // Room auth token is invalid
+    emit("invalidRoomAuthToken");
+  } else if (event.data.type === HTTP_ROOM_REQUIRE_CODE) {
+    // Forbidden, require access code
+    emit("invalidRoomAuthToken");
+  } else if (event.data.type === HTTP_ROOM_FILE_FORBIDDEN) {
+    // Forbidden, not allowed to view file
+    toast.error(t("rooms.flash.file_forbidden"));
+    EventBus.emit(EVENT_FORBIDDEN);
```
samuelwei, line 460:
```suggestion
    EventBus.emit(EVENT_FORBIDDEN);
    loadData();
```

Maybe also reload file list, in case the permissions to view the files have changed (Download is no longer allowed)

app/Http/Resources/Room.php
```diff
@@ -53,15 +53,15 @@ public function getDetails($latestMeeting)
         }
 
         return [
-            'username' => $this->when(! empty($this->token), ! empty($this->token) ? $this->token->fullname : null),
+            'username' => $this->when(! empty($this->personalizedLink), ! empty($this->personalizedLink) ? $this->personalizedLink->fullname : null),
```
samuelwei:
```suggestion
            'username' => $this->when(! empty($this->personalizedLink), $this->personalizedLink?->fullname),
```

app/Http/Controllers/api/v1/RoomPersonalizedLinkController.php
```diff
@@ -75,64 +75,64 @@ public function index(Room $room, Request $request)
...
-        $token->role = $request->role;
-        $room->tokens()->save($token);
+        $link = new RoomPersonalizedLink;
+        $link->firstname = $request->firstname;
+        $link->lastname = $request->lastname;
+        $link->role = $request->role;
+        $room->personalizedLinks()->save($link);
 
-        Log::info('Created new room token for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $token->role->label(), 'name' => $token->firstname.' '.$token->lastname]);
+        Log::info('Created new personalized room link for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $link->role->label(), 'name' => $link->firstname.' '.$link->lastname]);
```
samuelwei:
```suggestion
        Log::info('Created new personalized room link for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $link->role->label(), 'name' => $link->fullname]);
```

app/Http/Controllers/api/v1/RoomPersonalizedLinkController.php
```diff
@@ -75,64 +75,64 @@ public function index(Room $room, Request $request)
...
-        $token->lastname = $request->lastname;
-        $token->role = $request->role;
-        $token->save();
+        $link->firstname = $request->firstname;
+        $link->lastname = $request->lastname;
+        $link->role = $request->role;
+        $link->save();
 
-        Log::info('Updated room token for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $token->role->label(), 'name' => $token->firstname.' '.$token->lastname]);
+        Log::info('Updated personalized room link for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $link->role->label(), 'name' => $link->firstname.' '.$link->lastname]);
```
samuelwei:
```suggestion
        Log::info('Updated personalized room link for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $link->role->label(), 'name' => $link->fullname]);
```

app/Http/Controllers/api/v1/RoomPersonalizedLinkController.php
```diff
@@ -75,64 +75,64 @@ public function index(Room $room, Request $request)
...
-            abort(404, __('app.errors.token_not_found'));
+        if (! $link->room->is($room)) {
+            abort(404, __('app.errors.personalized_link_not_found'));
         }
 
-        $token->delete();
+        $link->delete();
 
-        Log::info('Removed room token for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $token->role->label(), 'name' => $token->firstname.' '.$token->lastname]);
+        Log::info('Removed personalized room link for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $link->role->label(), 'name' => $link->firstname.' '.$link->lastname]);
```
samuelwei:
```suggestion
        Log::info('Removed personalized room link for guest {name} with the role {role} for room {room}', ['room' => $room->getLogLabel(), 'role' => $link->role->label(), 'name' => $link->fullname]);
```

app/Services/RoomFileService.php
samuelwei: This method is now only used in the `MeetingService`. For the download url generation in the HTTP Resources we do not use this method. I therefor suggest removing this method and creating the download url directly inside of the MeetingService. 

The current solution would also result in the start meeting request to exit with 404 in case a file doesn't exists

database/migrations/2025_11_17_075446_create_room_auth_tokens_table.php
```diff
@@ -0,0 +1,41 @@
...
+            $table->integer('type');
+            $table->string('room_id');
+            $table->foreign('room_id')
+                ->references('id')->on('rooms')
+                ->onDelete('cascade');
+            $table->foreignId('room_personalized_link_id')
+                ->nullable()
+                ->constrained()
+                ->onDelete('cascade');
+            $table->timestamps();
```
samuelwei: Do we need or use the timestamps ?

resources/js/components/RoomJoinButton.vue
samuelwei: Adjust comment

resources/js/components/RoomMembershipButton.vue
```diff
@@ -100,10 +101,17 @@ function joinMembership() {
   isLoadingAction.value = true;
 
   // Join room as member, send access code if needed
```
samuelwei: adjust comment

resources/js/views/RoomsView.vue
```diff
@@ -536,6 +556,23 @@ function reload() {
     });
 }
 
+/**
+ * Reload room view, authenticate if token is provided
+ */
+function reloadRoomsView() {
+  if (props.token && !roomAuthToken.value) {
+    authenticate(ROOM_AUTH_TOKEN_TYPE_PERSONALIZED_LINK, props.token).then(
```
samuelwei, line 555: Why is this needed ?

samuelwei (changes requested):

database/migrations/2025_11_17_075446_create_room_auth_tokens_table.php
Sabr1n4W: To my knowledge the timestamps are not used anywhere and not necessarily needed here.

I think I added them because they were also added in the older branch and forgot to check if we still need them. 

If we remove them, we’ll need to update the model accordingly.

resources/js/views/RoomsView.vue
Sabr1n4W: This is needed for the following case:
* User tries to access room with a personalized room link but auth request that would return the room auth token fails with e.g. 500 error code.
* This results in `props.token`  being set and `roomAuthToken.value` being null 

After that the reload button is shown and can be clicked by the user to retry accessing the room.

When retrying to access the room it is necessary (for users that try to access using a personalized room link) to send the auth request again to retrieve a valid room auth token before trying to reload / load the room.

The frontend test `"room auth with personalized link errors"` in the file `RoomsViewGeneral.cy.js` contains checks that simulate this case (ll. 1362 -  1438)

This was fixed after a comment from coderabbit (https://github.com/THM-Health/PILOS/pull/2726#discussion_r2753674557) in the commits https://github.com/THM-Health/PILOS/pull/2726/commits/4c001ec56ba45ebf9412fed4ec31b905ee83e2bd and https://github.com/THM-Health/PILOS/pull/2726/changes/f48603600d4abc1e2f84bea457d58ead9de1a824. _(Tests were added and improved alongside other tests in other commits)_

Only reloading / loading the room would not be enough because the user would have no room auth token after the auth request failed and therefore the reload / load room request would fail in most cases.
If the room is accessible for guests without access code the request would be successful but the personalized link would not be used.

_An alternative for this would be to reload the whole page instead of only reloading the necessary requests._

Sabr1n4W: `props.token` is the personalized room link token provided by the route

Could maybe be renamed to make this clearer

resources/js/components/RoomTabFiles.vue
Sabr1n4W: I agree that reloading the file list, in case the permissions to view the files have changed (Download is no longer allowed) would be a good idea.

But I am currently not sure if it is that simple or if it could lead to problems with other reloads of the file list.

One scenario where this would result in several reloads is the following (not sure if there are others):
* User was a co-owner of the room but role was changed to something else while the user is on the file list
* The user tries to download a file which is not downloadable anymore because of the role change (previously had access to it as a co-owner but now is not able to download the file anymore -> error is triggered)
* This would trigger several reloads of the file list (with the changes above):
   * One because of the error handling
   * Another because of  `onRoomHasChanged` (user role changed from co-owner to something else)
   
Personally, I think reloading the file list several times in some cases would be better than not reloading when permissions to view the files have changed, but a solution that avoids triggering multiple reloads would be preferable.

database/migrations/2025_11_16_075447_add_room_personalized_links_table_indexes.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('room_personalized_links', function (Blueprint $table) {
+            $table->dropColumn(['id']);
+            $table->dropForeign(['room_id']);
+            $table->dropUnique(['token']);
```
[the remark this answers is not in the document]
samuelwei: Working with MariaDB 11.7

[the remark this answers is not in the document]
samuelwei: I tested and it works fine (as is) with MariaDB 11.7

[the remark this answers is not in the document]
samuelwei: No, Laravel doesn't drop the primary key. These are the raw statements from the DB logs:
```sql
alter table `room_personalized_links` drop `id`
alter table `room_personalized_links` drop foreign key `room_personalized_links_room_id_foreign`
alter table `room_personalized_links` drop index `room_personalized_links_token_unique`
```

PR #2799, opened 2026-02-07
Add ability to share rooms with roles (groups)

### LINKED ISSUE #2190: Feature Request: Share room with group
For us it would be extrem helpfull when we could add room permission to a group instead of a specified user.

so if user is in group "students" this user see all rooms assigned to to this group.

do you think something like that is possible?

currently user must loggin first via ldap tha a account will created, after that we can assign this user to rooms and than user see his rooms.

Thx!

Erik

hassanalitamam:
## Summary

Implements Feature Request #2190 — allows room owners to share rooms with entire roles (groups) instead of only individual users.

- **New `room_role` pivot table** following the existing `room_user` pattern
- **Backend API** with full CRUD for role-based room membership (add/edit/remove roles)
- **Room model updates** — `isMember()`, `isModerator()`, `isCoOwner()`, `getRole()` now check both individual and role-based membership, with highest role winning
- **Room listing** — "Shared rooms" filter includes rooms shared via roles
- **Frontend** — New role members section in the Members tab with search, filter, and pagination
- **Translations** — en, de, de-gender, fr, fa

### API Endpoints
| Method | Route | Description |
|--------|-------|-------------|
| GET | `rooms/{room}/role-member` | List role members |
| POST | `rooms/{room}/role-member` | Add role to room |
| PUT | `rooms/{room}/role-member/{role}` | Update role permission |
| DELETE | `rooms/{room}/role-member/{role}` | Remove role from room |

### Precedence
When a user has both individual and role-based membership, the highest permission level wins.

Closes #2190

## Test plan
- [ ] Run `php artisan migrate` — verify `room_role` table is created
- [ ] Add a role to a room — verify users in that role see the room in "Shared rooms"
- [ ] Test permission levels: participant, moderator, co-owner via role
- [ ] Test precedence: user with individual CO_OWNER + role-based USER still shows as CO_OWNER
- [ ] Test removing a role — verify users lose access
- [ ] Run `php artisan test` — all existing tests pass
- [ ] Run `npm run build` — frontend builds without errors

## Summary by CodeRabbit

* **New Features**
  * Assign roles as room members with permission levels (participant, moderator, co-owner)
  * New UI to list, search, filter, sort, add, edit and remove role-members (modals and action buttons)
  * New API endpoints to manage room role memberships

* **Behavior Changes**
  * Role-based membership is now included in room membership queries and permission checks

* **Documentation / Localization**
  * Added translation strings for role-member UI and validation messages


samuelwei:
Thank you for your PR!

Here are a few notes regarding this PR

1. We will not merge this at the moment, as we have the following PR that should be merged first:
- Refactoring the room authentication (Nearly complete, entering review and intense testing: #2726)

2. We already have several ways the share a room:
- Via link, internal only without access code
- Via link, internal only with access code
- Via link, public without access code
- Via link, public with access code
- With a single user
- With a known guest user via a personalized link

In addition we have to take care of the roles the user get's assigned for each of these options.
Managing this is currently split between the settings and the two tabs "Room members" and "Personalized room links". That makes it hard for the users to get an overview of all the available options to "share" a room.

Instead of just adding a second table on the "Room members" Tab we should instead work a broader concept.

3. I'm not sure if "Roles" is the correct association. I think a dedicated (new) model "Groups" would be a better fit. Mapping to a group could be the OU in LDAP, subdomain/domain in SSO, etc.

In that context it might make sense to organise groups in a tree. Sharing a room with a higher leaf (group) in the tree shares it with all leafs below. 

4. I think about replacing the direct sharing of a room with an invite system, so users are not added to rooms they don't want to be part of to prevent misuse. Unsure how this would look in a group share scenario. How can we prevent misuse? Do we need special permissions to allow users to share with a group?

PR #3147, opened 2026-05-18
Add unexpected toast message detection to Cypress tests

Sabr1n4W:
# Fixes \#

## Type

- Bugfix
- Feature
- Documentation
- **Refactoring** (e.g. Style updates, **Test implementation**, etc.)
- Other (please describe):

## Checklist

- [ ] Code updated to current develop branch head
- [ ] Passes CI checks
- [ ] Is a part of an issue
- [ ] Changelog is updated

## Changes

- Add check for unexpected toast messages as last check for all tests by overriding `it`
- Add same check for unexpected toast messages before visiting / reloading a page by overriding `cy.visit()` and `cy.reload()`
- Skip the unexpected toast check when the test switched origins (e.g. through simulating opening a link)
- Add missing explicit toast checks where needed

## Summary by CodeRabbit

* **Tests**
  * Added assertions to verify success and unauthorized toast notifications in login and admin access flows.
  * Added a final-state check that fails if unexpected toast messages remain at test end.
  * Updated the test harness to run the final-state check around navigation/reload actions so notifications are consistently validated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/THM-Health/PILOS/pull/3147?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


tests/Frontend/support/commands/commandOverrides.js
```diff
@@ -0,0 +1,53 @@
...
+ * Override for reload
+ * Checks final state before performing the reload, to make sure that unexpected errors that may have happened previously are caught.
+ */
+Cypress.Commands.overwrite("reload", (originalFn, ...args) => {
+  return cy
+    .wrap(null, { log: false })
+    .then(() => cy.checkFinalState())
+    .then(() => cy.log("Reload"))
+    .then(() => originalFn(...args));
+});
```
Sabr1n4W: The cypress logs are changed quite a bit because of the overrides.

e.g.
Before the overrides:
```
visit $url
```

After the overrides:
```
log Visit $url
- then $url
```

This is the case because the `originalFn` is called inside a `then` and therefore not logged as e.g. `visit` anymore but as `then` (I am not sure why this is the case but didn't find any way to "fix" this). 
The wrapping and calling `originalFn` is necessary because of the mix between calling a cypress custom command and calling `originalFn` which is not possible without the wrapping.

It would be possible to suppress the logs triggered by `originalFn` (the `-then` log) using `log: false` and also to partially "rebuild" a similar looking log entry using [Cypress.log](https://docs.cypress.io/api/cypress-api/cypress-log) and through this get log entries that mostly look like the ones before.

Even though this would be possible I would suggests to not suppress the logs triggered by the `originalFn` because they can contain some additional information which could be potentially helpful for debugging. Because of that I would also not try to "rebuild" a similar looking log entry because it could be misleading and confusing why this log entry does not contain all the additional information contained in the original log entry. 


Therefore I would suggest using the current log strategy.

Sabr1n4W:
There are still some explicit checks for missing toast messages at the end of tests, for example in `Login.cy.js` and `RoomsViewGeneral.cy.js`.

These checks would also be covered by the added checks for unexpected toast messages and could therefore be removed.

I currently kept them for now because in these tests the absence of a toast is part of the tested behavior, not just a generic cleanup/final-state condition.

However, keeping only some explicit missing toast message checks at the end of tests may also be a bit confusing because other tests may rely only on `cy.checkFinalState()` for similar expectations.

---

Decided to keep the explicit checks for missing toast messages at the end of tests

samuelwei (approved):

PR #3275, opened 2026-07-01
Refactor flow entering guest name and guest access

### LINKED ISSUE #2450: UX: Flow entering guest name
Currently guest users are asked their name when they click on "Start"/"Join" inside a modal

Image

sometimes also with additional consent to attendance logging, recording, streaming, etc.

Image

Until clicking on start/join some users might not realise, that they are not logged in to their user account.
Some users are also confused why their name and the consent is asked in the same step.

Sabr1n4W:
# Fixes \#2450

## Type

- Bugfix
- Feature
- Documentation
- Other (please describe):

## Checklist

- [x] Passes CI checks
- [x] Is a part of an issue
- [x] Changelog is updated

## Changes

- Move guest name input from join dialog to room access overlay
- Add possibility to remember the guest name for future visits
- Add possibility to pass access code and personalized link as hash params
- Include the room access code in room share links as a hash parameter
- Move personalized link tokens from route parameters to hash parameters
- Store access code and personalized link tokens in sessionStorage to preserve room access across page reloads
- Clear room tab / user tab hash when an invalid hash was provided and falling back to first / default tab

## Summary by CodeRabbit

* **New Features**
  * Introduced a room access overlay for guests with “continue as guest,” login-as-user, optional “remember name,” and participant-name editing.
  * Added a new participant name validation endpoint used by the room join flow.
  * Updated room/sharing links to include access-code/personalized-link details via URL hash.
* **Bug Fixes**
  * Improved resilience for invalid room/user/tab links by clearing bad hashes and tightening fallback behavior.
  * Preserved guest access state (including saved names/codes) across reloads.
* **Tests**
  * Expanded backend validation tests and refreshed/added Cypress E2E coverage for guest and access-code flows.
* **Chores**
  * Updated changelog and UI text/localization; minor layout spacing adjustments.


routes/api.php
```diff
@@ -175,6 +176,10 @@
         });
     });
 
+    Route::post('participantName/check', function (ValidateParticipantNameRequest $request) {
+        return response()->noContent();
+    })->name('participantName.check');
+
```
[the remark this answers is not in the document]
samuelwei: See https://github.com/THM-Health/PILOS/blob/develop/app/Providers/RouteServiceProvider.php#L52. All API routes have a base rate limit, some routes have a different more strict limit

resources/js/components/RoomAccessOverlay.vue
```diff
@@ -0,0 +1,248 @@
...
+    :pt="{ header: { class: 'flex justify-center' } }"
+    data-test="room-access-overlay"
+  >
+    <template #header>
+      <Badge
+        severity="danger"
+        class="-mt-8 flex !h-16 !w-16 items-center justify-center rounded-full"
+      >
+        <i class="fa-solid fa-lock text-2xl text-white"></i>
+      </Badge>
```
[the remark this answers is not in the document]
samuelwei: Will be done in an other PR, same issue in many other files

resources/js/views/RoomsView.vue
```diff
@@ -181,21 +114,66 @@
...
             </template>
             <template #content>
-              <div v-if="room.can_start && room.room_type_invalid" class="mb-4">
-                <Message
+              <div class="flex flex-col gap-2">
+                <InlineNote
+                  v-if="room.last_meeting?.detached"
                   severity="warn"
-                  icon="fa-solid fa-unlink"
+                  :class="
```
samuelwei, line 100: Unsure about this. Could be move this to the parent div or increase the gap there?

resources/js/components/RoomJoinButton.vue
```diff
@@ -379,7 +363,7 @@ function getJoinUrl() {
   const config = {
     method: "post",
     data: {
-      name: !requiresGuestName.value ? null : name.value,
+      name: props.participantName ? props.participantName : null,
```
samuelwei:
```suggestion
      name: props.participantName,
```

resources/js/components/RoomAccessOverlay.vue
```diff
@@ -0,0 +1,230 @@
+<template>
+  <Card
+    style="width: 500px; max-width: 90vw"
+    :pt="{ header: { class: 'flex justify-center' } }"
+    data-test="room-access-overlay"
+  >
+    <template #header>
+      <Badge
```
samuelwei, line 8:
```suggestion
      <Badge
        v-if="!room.authenticated"
```

IMO the lock icon only makes sense if the room requires an access code

resources/js/views/RoomsView.vue
samuelwei, line 290: We could add an eventListener to detect user-triggered changes to the hash.
If the hash is changed via hashParams the event is not triggered (https://developer.mozilla.org/en-US/docs/Web/API/Window/hashchange_event)

Currently if the page is already loaded and we get an auth error, e.g. in the meantime an access code has been added or changes, we cannot copy the new url and paste it in the url

```js
window.addEventListener("hashchange", () => {
    console.log(hashParams.accessCode);
});
```

samuelwei (commented):
RoomView.vue Review still pending due to its complexity

resources/js/components/RoomAccessOverlay.vue
```diff
@@ -0,0 +1,230 @@
...
+            v-model="accessCodeInput"
+            autofocus
+            :mask="room.legacy_code ? '******' : '999-999-999'"
+            :placeholder="room.legacy_code ? '123abc' : '123-456-789'"
+            :invalid="
+              accessCodeInvalid || formErrors.fieldInvalid('access_code')
+            "
+            :disabled="authThrottledFor > 0"
+            class="text-center"
+            @keydown.enter="submit"
```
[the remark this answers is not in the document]
samuelwei: Confirmed we need @keydown.enter="submit" as the default Enter inside the form is not triggered. Could be caused by a keydown listener for InputMask in PrimeVue. I'm unsure why this an issue as the event listener in primevue doesnt prevent the default (https://github.com/primefaces/primevue/blob/master/packages/primevue/src/inputmask/InputMask.vue#L154) not without this extra @keydown.enter the form is not submitted

resources/js/views/RoomsView.vue
samuelwei:
```suggestion
  tokenInvalid.value = true;
  sessionStorage.removeItem("roomPersonalizedLink_" + props.id);
```

Unsure about the behavior when entering an invalid personalized room link.

Before this PR the user could paste the regular room url to load without a personalized link.

With the PR the user cannot paste the regular room url, the invalid personalized link is persisted in the local storage.
The only option to successfully load the room is either pasting a new valid personalized link or a link with an access code.

Sabr1n4W: Not persisting the personalized link in the session storage could lead to the following problem (because the hash is cleared):
* User tries to enter the room with an invalid personalized link
* Error message is shown that the personalized link is invalid
* User reloads the whole page => User enters the room as "normal" guest instead of the personalized link without noticing it (especially if the room does not have an access code)

Because of that the personalized link is currently persisted to make sure that the user does not "change" to the "normal" guest access without noticing it / intending to do so if I remember it correctly

An alternative approach could maybe be to keep the personalized link in the hash and only clear it after the auth request was successfull

Another possible solution I thought about was maybe offering an explicit way to change to the "normal" guest access on the error page

resources/js/views/RoomsView.vue
```diff
@@ -721,6 +730,139 @@ function authenticate(type, codeOrToken) {
...
+
+    if (savedAccessCode) {
+      accessCodeInput.value = savedAccessCode;
+    }
+  }
+
+  if (!authStore.isAuthenticated) {
+    await loadSavedGuestName();
+  }
+}
```
samuelwei, line 778:
```suggestion
async function loadSavedAccessParameters() {
  if (!authStore.isAuthenticated) {
    await loadSavedGuestName();
  }

  // Load personalized link from hash params
  if (hashParams.personalizedLink) {
    personalizedLink.value = hashParams.personalizedLink;

    sessionStorage.setItem(
      "roomPersonalizedLink_" + props.id,
      hashParams.personalizedLink,
    );

    // Clear hash params
    await nextTick();
    hashParams.personalizedLink = null;

    // Treat a personalized link from the hash as the selected credential
    // discard any competing hash access code so it cannot win on a later reload when
    // the personalized link is only stored in session storage
    hashParams.accessCode = null;

    return;
  }

  // Load Access Parameters stored in hash params
  if (hashParams.accessCode) {
    accessCodeInput.value = hashParams.accessCode;
    // Set access code param in session storage to make sure that access code
    // stays set if the user decides to log in (navigates to login page)
    sessionStorage.setItem("roomAccessCode_" + props.id, hashParams.accessCode);

    // Treat an access code from the hash as the selected credential
    // discard any saved personalized link so it cannot win on a later reload when
    // the access code is only stored in session storage
    sessionStorage.removeItem("roomPersonalizedLink_" + props.id);

    // Clear hash params
    await nextTick();
    hashParams.accessCode = null;

    return;
  }

  // Load Access Parameters stored in session storage only if parameters are not already set by hash params
  if (accessCodeInput.value === "") {
    const savedPersonalizedLink = sessionStorage.getItem(
      "roomPersonalizedLink_" + props.id,
    );
    if (savedPersonalizedLink) {
      if (authStore.isAuthenticated) {
        // User is authenticated and not allowed to use personalized link, remove it from session storage
        // and continue without loading the existing personalized link
        sessionStorage.removeItem("roomPersonalizedLink_" + props.id);
      } else {
        personalizedLink.value = savedPersonalizedLink;
        return;
      }
  const savedPersonalizedLink = sessionStorage.getItem(
    "roomPersonalizedLink_" + props.id,
  );
  if (savedPersonalizedLink) {
    if (authStore.isAuthenticated) {
      // User is authenticated and not allowed to use personalized link, remove it from session storage
      // and continue without loading the existing personalized link
      sessionStorage.removeItem("roomPersonalizedLink_" + props.id);
    } else {
      personalizedLink.value = savedPersonalizedLink;
    }

    const savedAccessCode = sessionStorage.getItem(
      "roomAccessCode_" + props.id,
    );

    if (savedAccessCode) {
      accessCodeInput.value = savedAccessCode;
    }
    return;
  }

  if (!authStore.isAuthenticated) {
    await loadSavedGuestName();
  const savedAccessCode = sessionStorage.getItem("roomAccessCode_" + props.id);
  if (savedAccessCode) {
    accessCodeInput.value = savedAccessCode;
  }
}
```

- Move loadSavedGuestName to the top, always load the name from storage
- Return after each step, removed complex nested code flows

resources/js/views/RoomsView.vue
samuelwei: Fair point, maybe adding a new route to redirect users to? I'll try it

